PR #7, opened 2021-11-22
feat: intro metric-adapter to exports prediction pod metrics

qmhu:
Change list in this pr:
- metric adapter cmd
- metric adapter deploy yaml
- metric adapter build scripts
- use custom metric instead of external metric
- use origin hpa metric and prediction metric to build hpa metrics
- support utilization metric target


pkg/controller/hpa/hpa.go
```diff
@@ -73,25 +76,29 @@ func (p *AdvancedHPAController) CreateHPA(ctx context.Context, ahpa *autoscaling
...
 	hpa := &autoscalingv2.HorizontalPodAutoscaler{
 		ObjectMeta: metav1.ObjectMeta{
-			Namespace: ahpa.Namespace, // the same namespace to ahpa
+			Namespace: ahpa.Namespace, // the same namespace to advancedhpa
 			Name:      name,
 			Labels: map[string]string{
 				"app.kubernetes.io/name":                      name,
 				"app.kubernetes.io/part-of":                   ahpa.Name,
-				"app.kubernetes.io/managed-by":                "advanced-hpa-controller",
+				"app.kubernetes.io/managed-by":                known.AdvancedHorizontalPodAutoscalerManagedBy,
```
mfanjie, line 93: we should use ownerReference to HPA so that we don't need to delete them in controller code.

pkg/known/const.go
```diff
@@ -4,3 +4,8 @@ const (
 	CraneSystemNamespace = "crane-system"
 	AutoscalingFinalizer = "autoscaling.crane.io/finalizer"
```
mfanjie, line 5: this comment was missed in last pr, if we use ownerRef, seems this finalizer is not required?

cmd/metric-adapter/main.go
```diff
@@ -0,0 +1,110 @@
...
+	logs.InitLogs()
+	defer logs.FlushLogs()
+
+	cmd := &MetricAdapter{}
+
+	cmd.OpenAPIConfig = genericapiserver.DefaultOpenAPIConfig(generatedopenapi.GetOpenAPIDefinitions, openapinamer.NewDefinitionNamer(scheme))
+	cmd.OpenAPIConfig.Info.Title = "crane-metric-adapter"
+	cmd.OpenAPIConfig.Info.Version = "1.0.0"
+
+	cmd.Flags().StringVar(&cmd.Message, "msg", "starting adapter...", "startup message")
```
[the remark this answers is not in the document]
qmhu: Agreed,  all logs should start with Capital letter

pkg/controller/hpa/hpa.go
```diff
@@ -4,16 +4,19 @@ import (
 	"context"
 	"fmt"
 
+	"k8s.io/klog/v2"
+
```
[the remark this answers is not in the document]
qmhu: this is from goimports ... anyway i will combined them

pkg/controller/hpa/hpa.go
qmhu: ok, I will remove the deletion code and for the finalizer I want to keep it ,because if we want to do some clean work other than delete hpa and timeseries in future, the finalizer is useful for that.

pkg/known/const.go
qmhu: the same above

pkg/controller/hpa/hpa.go
mfanjie: finalizer introduce some difficulties to beginners, as we are open sourcing the project, if someone tear down crane-system namespace first, then the manager will be down, and all apha will be stuck on deletion.

PR #12, opened 2021-11-26
update status and implement scale target

qmhu:
1. update ahpa status based on reconciler loop
2. implement scale strategy Manual


pkg/controller/hpa/advanced_hpa_controller.go
```diff
@@ -43,25 +46,92 @@ func (p *AdvancedHPAController) Reconcile(ctx context.Context, req ctrl.Request)
...
 	}
 
-	// todo: update status
-	// autoscalingStatusOrigin := autoscaler.Status.DeepCopy()
+	setCondition(newStatus, autoscalingapi.Ready, metav1.ConditionTrue, "AdvancedHorizontalPodAutoscalerReady", "Advanced HPA is ready")
+	return ctrl.Result{}, p.UpdateStatus(ctx, ahpa, newStatus)
+}
+
+func (p *AdvancedHPAController) UpdateStatus(ctx context.Context, ahpa *autoscalingapi.AdvancedHorizontalPodAutoscaler, newStatus *autoscalingapi.AdvancedHorizontalPodAutoscalerStatus) error {
+	if !equality.Semantic.DeepEqual(&ahpa.Status, newStatus) {
```
mfanjie, line 120: not a comment but a question, what's the different of this function and reflect.DeepEqual?

pkg/controller/hpa/advanced_hpa_controller.go
```diff
@@ -43,25 +46,92 @@ func (p *AdvancedHPAController) Reconcile(ctx context.Context, req ctrl.Request)
...
+		p.Log.Info("AdvancedHorizontalPodAutoscaler status should be updated", "currentStatus", &ahpa.Status, "newStatus", newStatus)
```
mfanjie, line 121: can we always set log level for info?
and if you are dumping the entire status, the level should be at least 4.

pkg/controller/hpa/advanced_hpa_controller.go
```diff
@@ -43,25 +46,92 @@ func (p *AdvancedHPAController) Reconcile(ctx context.Context, req ctrl.Request)
...
+
+		ahpa.Status = *newStatus
+		err := p.Status().Update(ctx, ahpa)
+		if err != nil {
+			p.Recorder.Event(ahpa, v1.EventTypeNormal, "FailedUpdateStatus", err.Error())
+			p.Log.Error(err, "Failed to update status", "advanced-hpa", klog.KObj(ahpa))
+			return err
+		}
+
+		p.Log.Info("Update AdvancedHorizontalPodAutoscaler status successful", "advanced-hpa", klog.KObj(ahpa))
```
mfanjie, line 131: do we have a short name `ahpa` for the object, if yes, we can simply dump ahpa.

pkg/controller/hpa/advanced_hpa_controller.go
qmhu: reflect.DeepEqual will not necessarily return true for semantically equal quantity objects. That is the purpose of the semantic equality DeepEqual method (apiequality.Semantic.DeepEqual, see https://godoc.org/k8s.io/apimachinery/pkg/api/equality)

pkg/controller/hpa/advanced_hpa_controller.go
qmhu: klog.KObj will format the object like "{namespace}/{name}".

PR #15, opened 2021-12-01
move webhook from api to crane

mfanjie:


mfanjie:
as pod group prediciton is being deprecated, this pr is no longer needed

PR #24, opened 2021-12-03
Feature/time series prediction controller

kitianFresh:


cmd/craned/app/manager.go
```diff
@@ -84,6 +93,60 @@ func Run(ctx context.Context, opts *options.Options) error {
 		return err
 	}
 
+	var dataSource providers.Interface
```
qmhu, line 96: would you please move it to  function initializationControllers

pkg/controller/tsp/status.go
```diff
@@ -0,0 +1,382 @@
...
+		tsPrediction := &predictions[i]
+		newStatus := tsPrediction.Status.DeepCopy()
+		key := GetTimeSeriesPredictionKey(tsPrediction)
+		tc.Logger.Info("diff asw and dsw", "key", key)
+		if err := tc.Client.Get(context.TODO(), client.ObjectKey{Name: tsPrediction.Name, Namespace: tsPrediction.Namespace}, tsPrediction); err != nil {
+			// If the prediction does not exist any more, we delete the prediction data from the map.
+			if apierrors.IsNotFound(err) {
+				tc.tsPredictionMap.Delete(key)
+			}
+			tc.Logger.Error(err, "", key, err)
```
qmhu, line 45: please add error message here

pkg/controller/tsp/status.go
```diff
@@ -0,0 +1,382 @@
...
+		// force predict and update the status
+		if len(warnings) > 0 {
+			start := time.Now()
+			// double the time to predict so that crd consumer always see time series range [now, now + PredictionWindowSeconds] in PredictionWindowSeconds window
+			end := start.Add(time.Duration(tsPrediction.Spec.PredictionWindowSeconds) * time.Second * 2)
+
+			predictedData, err := tc.doPredict(tsPrediction, start, end)
+			if err != nil {
+				tc.Recorder.Event(tsPrediction, v1.EventTypeWarning, "FailedPredict", err.Error())
+				tc.Logger.Error(err, "failed to doPredict")
```
qmhu, line 63: Please Uppercase the first word in logging for the whole commit

pkg/controller/tsp/status.go
```diff
@@ -0,0 +1,382 @@
...
+		} else if metric.MetricSelector != nil {
+			//todo
+		}
+	}
+	return result, nil
+}
+
+func (tc *Controller) UpdateStatus(ctx context.Context, tsPrediction *v1alpha1.TimeSeriesPrediction, newStatus *v1alpha1.TimeSeriesPredictionStatus) error {
+	if !equality.Semantic.DeepEqual(&tsPrediction.Status, newStatus) {
+		tc.Logger.Info("status should be updated", "currentStatus", &tsPrediction.Status, "newStatus", newStatus)
```
qmhu, line 284: We can use logger level 4 here for status diff

pkg/controller/tsp/status.go
```diff
@@ -0,0 +1,382 @@
...
+
+		tsPrediction.Status = *newStatus
+		err := tc.Status().Update(ctx, tsPrediction)
+		if err != nil {
+			tc.Recorder.Event(tsPrediction, v1.EventTypeNormal, "FailedUpdateStatus", err.Error())
+			tc.Logger.Error(err, "Failed to update status", "time-series-prediction", klog.KObj(tsPrediction))
```
qmhu, line 290: s/time-series-prediction/TimeSeriesPrediction/g

pkg/controller/tsp/status.go
```diff
@@ -0,0 +1,382 @@
...
+	}
+	return list
+}
+
+// UpdateTimeSeriesPredictionCondition updates existing timeseriesprediction condition or creates a new one. Sets LastTransitionTime to now if the
+// status has changed.
+// Returns true if pod condition has changed or has been added.
+func UpdateTimeSeriesPredictionCondition(status *v1alpha1.TimeSeriesPredictionStatus, condition *metav1.Condition) bool {
+	condition.LastTransitionTime = metav1.Now()
+	// Try to find this podgroupprediction condition.
```
qmhu, line 324: podgroupprediction ?

pkg/controller/tsp/status.go
```diff
@@ -0,0 +1,382 @@
...
+	conditionIndex, oldCondition := GetTimeSeriesPredictionCondition(status, condition.Type)
+
+	if oldCondition == nil {
```
qmhu, line 327: Can we use `equality.Semantic.DeepEqual` here ?

pkg/controller/tsp/status.go
```diff
@@ -0,0 +1,382 @@
...
+
+	"github.com/gocrane/crane/pkg/common"
+	"github.com/gocrane/crane/pkg/known"
+	"github.com/gocrane/crane/pkg/prediction"
+)
+
+// Scan all TimeSeriesPredictions and update the status if it is needed,
+// update each time series prediction status window length is double of the spec.PredictionWindowSeconds.
+// check the actual state of world and decide if need to update the crd status, it is periodic check to meet updateStatusDelayQueue's flaw.
+func (tc *Controller) syncPredictionsStatus() error {
```
qmhu, line 28: what is the difference between syncPredictionsStatus and updateStatusDelayQueue? they all update tsp status

cmd/craned/app/manager.go
kitianFresh: because there has some other params, if so we should change the initializationControllers func params

qmhu: Add more params for initializationControllers is better

cmd/craned/app/options/options.go
```diff
@@ -44,4 +55,18 @@ func (o *Options) AddFlags(flags *pflag.FlagSet) {
 	flags.DurationVar(&o.LeaderElection.RenewDeadline.Duration, "lease-renew-period", 10*time.Second,
 		"Specifies the lease renew interval.")
 
+	flags.DurationVar(&o.PredictionUpdateFrequency, "prediction-update-frequency-duration", 30*time.Second,
+		"Specifies the update frequency of the prediction.")
+	flags.StringVar(&o.DataSource, "datasource", "prom", "data source of the predictor, prom, mock is available")
```
mfanjie, line 60: description start with Capital letters, keep same style with others

examples/tsp-workload-resource-dsp.yaml
```diff
@@ -0,0 +1,44 @@
+apiVersion: prediction.crane.io/v1alpha1
+kind: TimeSeriesPrediction
+metadata:
+  name: workload-resource-dsp
+  namespace: default
+spec:
+  predictionWindowSeconds: 3600
+  predictionMetrics:
+    - resourceIdentifier: workload-cpu
+      workloadRef:
```
mfanjie, line 10: I think it would be better if we can unify workloadRef and NodeRef and only one refer is allowed for one tsp, so that the resource can be kept under each resourceIdentifier, but the workloadRef should be something like
```
  TargetRef: (for workload)
        kind: Deployment
        name: dep-1-100m-500mib
        namespace: default
```
```
TargetRef: (for node)
        kind: Node
        name: dep-1-100m-500mib
```

pkg/controller/tsp/status.go
```diff
@@ -0,0 +1,385 @@
+package tsp
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"strings"
+	"time"
+
+	predconfig "github.com/gocrane/crane/pkg/prediction/config"
```
mfanjie, line 10: format import

cmd/craned/app/options/options.go
kitianFresh: ok

pkg/controller/tsp/status.go
kitianFresh: updateStatusDelayQueue must do predict when the time is ready based on the PredictionWindowSeconds in crd. when it update, next PredictionWindowSeconds it will enqueue again.

But this is not enough, when user change the PredictionWindowSeconds or change the algorithm, we should update it, periodic check it better. we can remove updateStatusDelayQueue if we just use periodic check.

examples/tsp-workload-resource-dsp.yaml
kitianFresh: > I think it would be better if we can unify workloadRef and NodeRef and only one refer is allowed for one tsp, so that the resource can be kept under each resourceIdentifier, but the workloadRef should be something like
> 
> ```
>   TargetRef: (for workload)
>         kind: Deployment
>         name: dep-1-100m-500mib
>         namespace: default
> ```
> 
> ```
> TargetRef: (for node)
>         kind: Node
>         name: dep-1-100m-500mib
> ```

OK, I will move the TargetRef to the tsp scope, and each tsp associate just only one TargetRef

PR #29, opened 2021-12-08
Recommendation controller

qmhu:


pkg/controller/recommendation/recommendation_controller.go
```diff
@@ -0,0 +1,101 @@
...
+	err := c.Client.Get(ctx, req.NamespacedName, recommendation)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	newStatus := recommendation.Status.DeepCopy()
+
+	recommender, err := recommend.NewRecommender(c.Client, c.RestMapper, c.ScaleClient, recommendation)
+	if err != nil {
+
```
kitianFresh, line 45: what to do for these err?

pkg/recommend/recommender.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	c.Scale = scale
+	c.RestMapping = mapping
+
+	unstructured := &unstructured.Unstructured{}
+	if err = kubeClient.Get(context.TODO(), client.ObjectKey{Namespace: recommendation.Namespace, Name: recommendation.Spec.TargetRef.Name}, unstructured); err != nil {
+		return nil, err
+	}
+
+	if recommendation.Spec.TargetRef.Kind == "Deployment" && mapping.GroupVersionKind.Group == "apps" {
```
kitianFresh, line 74: Maybe we can define these kind string in consts

pkg/controller/ehpa/substitute_controller.go
```diff
@@ -61,24 +60,13 @@ func (c *SubstituteController) Reconcile(ctx context.Context, req ctrl.Request)
 		c.Log.Info("Update Substitute status successful", "Substitute", klog.KObj(substitute))
 	}
 
-	// Rsync every 1 minute
+	// Rsync every 15 seconds
```
mfanjie, line 63: 1. is 15s too fast? if there are 5000 subs, then the controller will be busy on reconcile almost all the time.
2. can we make this a customize option with 1min default interval?

pkg/utils/scale.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	podList := &v1.PodList{}
+	err = kubeClient.List(context.TODO(), podList, opts...)
+	if err != nil {
+		return nil, err
+	}
+
+	return podList.Items, nil
+}
```
mfanjie, line 62: did you miss fmt?

cmd/craned/app/manager.go
```diff
@@ -9,16 +9,21 @@ import (
...
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/dynamic"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/scale"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 
 	"github.com/spf13/cobra"
 
+	analysisnapi "github.com/gocrane/api/analysis/v1alpha1"
```
mfanjie, line 21: is napi a typo?

pkg/controller/ehpa/substitute_controller.go
qmhu: The background is controller sync not only selectors here but also replicas, since HPA controller reconcile hpa every 15 seconds by default, so we may follow the 15 seconds here.
This can be a start options.

pkg/controller/recommendation/recommendation_controller.go
qmhu: addressed

PR #30, opened 2021-12-08
add webhook for tsp

mfanjie:
Add webhook framework to crane project, currently tsp mutating and validating is added, it will be easier to support more objects.
webhooks server requires tls support, so the existing craned deployment is updated with a secret volumeMount change.
the initial key/cert for the webhook is checked in, if new certs is needed, user can run `hack/scripts/gencerts.sh`


cmd/craned/app/manager.go
```diff
@@ -94,6 +94,13 @@ func Run(ctx context.Context, opts *options.Options) error {
 
 	return nil
 }
+func initializationWebhooks(mgr ctrl.Manager, opts *options.Options) {
+	log.Logger().Info(fmt.Sprintf("opts %v", opts))
+	if err := webhooks.SetupWebhookWithManager(mgr); err != nil {
+		log.Logger().Error(err, "unable to create webhook", "webhook", "SimpleStatefulset")
```
qmhu, line 100: a typo `SimpleStatefulset `?

qmhu (commented):
/LGTM

mfanjie:
rebased and resolved conflict

PR #34, opened 2021-12-10
refine tsp controller, fix some bug and remove some code

kitianFresh:


pkg/controller/tsp/tsp_controller.go
```diff
@@ -139,17 +105,18 @@ func (tc *Controller) syncTimeSeriesPrediction(prediction *predictionv1alph1.Tim
...
 }
 
 func NewMetricContext(prediction *predictionv1alph1.TimeSeriesPrediction) *predconfig.MetricContext {
 	c := &predconfig.MetricContext{
 		Namespace:  prediction.Namespace,
 		TargetKind: prediction.Spec.TargetRef.Kind,
+		Name:       prediction.Spec.TargetRef.Name,
 	}
-	if c.TargetKind != predconfig.TargetKindNode && prediction.Spec.TargetRef.Namespace != "" {
+	if strings.ToLower(c.TargetKind) != predconfig.TargetKindNode && prediction.Spec.TargetRef.Namespace != "" {
```
mfanjie, line 119: I see the actual const is Node, a toLower will make this as false all the time?
const TargetKindNode = "Node"

pkg/controller/tsp/status.go
```diff
@@ -22,217 +22,91 @@ import (
...
-		return err
+// check and update the status if it is needed, update each time series prediction status window length is double of the spec.PredictionWindowSeconds.
+// check the actual state of world and decide if need to update the crd status,
+// driven by time tick not by events, because time series prediction need to update the prediction window data to avoid the data is out of date.
+// NOTE: update period is better higher resolution than the algorithm sample interval, reduce the possibility of the data is out date.
+// but it is a final consistent system, so the data will be in date when next update reconcile in controller runtime.
+func (tc *Controller) syncPredictionStatus(ctx context.Context, tsPrediction *v1alpha1.TimeSeriesPrediction) (ctrl.Result, error) {
+	newStatus := tsPrediction.Status.DeepCopy()
+	key := GetTimeSeriesPredictionKey(tsPrediction)
+	tc.Logger.Info("SyncPredictionsStatus check asw and dsw", "key", key)
```
mfanjie, line 33: suggest to log level to all info

pkg/controller/tsp/tsp_controller.go
```diff
@@ -102,23 +77,14 @@ func (tc *Controller) SetupWithManager(mgr ctrl.Manager) error {
...
-		//newStatus := prediction.Status.DeepCopy()
-		//cond := &predictionv1alph1.TimeSeriesPredictionCondition{
-		//	Type:               predictionv1alph1.TimeSeriesPredictionConditionNotReady,
-		//	Status:             metav1.ConditionTrue,
-		//	LastProbeTime:      metav1.Now(),
-		//	LastTransitionTime: metav1.Now(),
-		//}
-		//UpdateTimeSeriesPredictionCondition(newStatus, cond)
 	} else {
 		if old, ok := last.(*predictionv1alph1.TimeSeriesPrediction); ok {
```
qmhu, line 89: which scene need to compare old prediction and new prediction here?

kitianFresh: delete the old configs, because the internal config is cached in memory in prediction

PR #35, opened 2021-12-11
Add CNP Controller & Fixed TSP Reconcile not found 

yekangming:
1. fixed reconcile not found object issue 
2. Unified Controller object is c 
3. Unified MetricsContext is mc
4. Add CNP Controller


pkg/controller/cnp/clusternodeprediction_controller.go
```diff
@@ -0,0 +1,114 @@
+package cnp
+
+import (
+	"context"
```
mfanjie, line 4: make system import another section

pkg/controller/cnp/clusternodeprediction_controller.go
```diff
@@ -0,0 +1,114 @@
...
+	predictionapi "github.com/gocrane/api/prediction/v1alpha1"
+)
+
+type ClusterNodePredictionController struct {
+	client.Client
+	Log         logr.Logger
+	Scheme      *runtime.Scheme
+	RestMapper  meta.RESTMapper
+	Recorder    record.EventRecorder
+	scaleClient scale.ScalesGetter
```
mfanjie, line 25: scaleClient is not needed here, please remove it.

pkg/controller/cnp/clusternodeprediction_controller.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+		if err := c.updateStatus(ctx, &cnp, status); err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, nil
+	}
+
+	// Mutate TimeSeriesPrediction according to the number of node list
+	var successCount int
+	for _, node := range nodeList.Items {
```
mfanjie, line 65: seems when never there is an error, the entire sync loop exit without updating the status? it's better if we can support partial success status, as the actual tsp is created already.

pkg/controller/cnp/clusternodeprediction_controller.go
```diff
@@ -0,0 +1,114 @@
...
+			return controllerutil.SetControllerReference(&cnp, &tsp, c.Scheme)
+		}); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	// Set the status, Then reconcile func end
+	status := predictionapi.ClusterNodePredictionStatus{
+		CurrentNumberCreated:successCount,
+		DesiredNumberCreated: len(nodeList.Items),
```
mfanjie, line 81: can we add condition here?

pkg/controller/cnp/clusternodeprediction_controller.go
```diff
@@ -0,0 +1,114 @@
...
+	Scheme      *runtime.Scheme
+	RestMapper  meta.RESTMapper
+	Recorder    record.EventRecorder
+	scaleClient scale.ScalesGetter
+	K8SVersion  *version.Version
+}
+
+func (c *ClusterNodePredictionController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	// Log for controller information
+	c.Log.Info("got", "cnp", req.NamespacedName)
```
mfanjie, line 31: cloud you please more logs/event. for main code branches (and set log level to 4 for debug info if there is any)

mfanjie (commented):
The overall logic is pretty clear @yekangming , thanks for the contribution
can you also consider the following case
1. a node watcher is needed because node can come and go, we need to create and delete tsp automatically.
2. when the template of cnp is changed, we might need to rollout the change to all tsp with a predefined policy.

the 2nd case can be in differnet pr as it is pretty big, can you consider 1 in this pr?

mfanjie:
and need to update go.mod to include latest api change, seems the build is failing.

PR #51, opened 2021-12-21
refine node resource controller

mfanjie:
add node resource controller, which behaves as the following:
1. watches all tsp
2. when tsp target ref is node, it list all nodes and find the target
3. it do nothing if tsp resource is not cpu or memory
4. it use capacity - predict result for the possible idle resource（max value of the prediction result)
5. when delta > 10%, it update node status with extended resource capacity

### DIFF
--- a/pkg/controller/noderesource/node_resource_controller.go
+++ b/pkg/controller/noderesource/node_resource_controller.go
@@ -1,0 +1,117 @@
+/*
+Copyright 2021.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package noderesource
+
+import (
+	"context"
+	"strconv"
+
+	"github.com/go-logr/logr"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/client-go/tools/record"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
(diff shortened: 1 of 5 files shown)


pkg/controller/noderesource/node_resource_controller.go
```diff
@@ -0,0 +1,171 @@
...
+				nextRecommendation = node.Status.Allocatable.Memory().Value() - maxUsage
+			default:
+				continue
+			}
+			extResourceName := fmt.Sprintf(ExtResourcePrefix, string(*resourceName))
+			resValue, exists := node.Status.Capacity[v1.ResourceName(extResourceName)]
+			if exists && float64(resValue.Value())/float64(nextRecommendation) <= 1+MinDeltaRatio {
+				continue
+			}
+			node.Status.Capacity[v1.ResourceName(extResourceName)] =
```
kitianFresh, line 160: The scheduler use node.Status.Allocatable to do scheduling,  we should update the node.Status.Allocatable too.

pkg/controller/noderesource/node_resource_controller.go
```diff
@@ -0,0 +1,171 @@
...
+			var nextRecommendation int64
+			switch *resourceName {
+			case v1.ResourceCPU:
+				nextRecommendation = node.Status.Allocatable.Cpu().Value() - maxUsage
+			case v1.ResourceMemory:
+				nextRecommendation = node.Status.Allocatable.Memory().Value() - maxUsage
+			default:
+				continue
+			}
+			extResourceName := fmt.Sprintf(ExtResourcePrefix, string(*resourceName))
```
kitianFresh, line 155: I think we should may have a requests resource amount threshold feature gate to decide when Node ext resource enabled, because when the total resource requests in the node is little, such as there are 32c64G capacity, but the node's total requests is only 1c2G, we do not need to expose the ext resource, because the static requests resource has not reach too high, it has enough requests quota to schedule pods.

pkg/controller/noderesource/node_resource_controller.go
mfanjie: done

pkg/controller/noderesource/node_resource_controller.go
mfanjie: I think this up to user, because the ext resource reallocation is based on crd, as long as user do not create tsp for small node, we won't update the ext-resource.
so we do not need to add such checks in our code, as it's hard to define what's it small from our perspective.

PR #52, opened 2021-12-21
Supports analyzing workloads in cluster scope by creating 'Analytics' in namespace 'crane-system'

yufeiyu:


pkg/controller/analytics/analytics_controller.go
```diff
@@ -30,6 +30,8 @@ import (
 	analysislister "github.com/gocrane/api/pkg/generated/listers/analysis/v1alpha1"
 )
 
+const CraneSystemNamespace = "crane-system"
```
kitianFresh, line 33: this const is in the package known

yufeiyu: resolved

PR #54, opened 2021-12-22
Bugfix/node resource recommendation less than zero

kitianFresh:


pkg/controller/noderesource/node_resource_controller.go
```diff
@@ -154,11 +155,22 @@ func (r *NodeResourceReconciler) BuildNodeStatus(tsp *predictionapi.TimeSeriesPr
...
 			}
+			// todo: maybe use float and milli value for cpu maxUsage
+			if nextRecommendation < 0 {
+				r.Log.Info("unexpected recommendation", "nodeName", node.Name, "maxUsage", maxUsage, "nextRecommendation", nextRecommendation)
+				continue
+			}
 			extResourceName := fmt.Sprintf(ExtResourcePrefix, string(*resourceName))
 			resValue, exists := node.Status.Capacity[v1.ResourceName(extResourceName)]
-			if exists && float64(resValue.Value())/float64(nextRecommendation) <= 1+MinDeltaRatio {
+			if !exists {
```
mfanjie, line 165: we should update the status when the key does not exist.

mfanjie:
fixed by https://github.com/gocrane/crane/pull/62

PR #70, opened 2021-12-30
refect deploy dir

qmhu:


README.md
```diff
@@ -64,3 +64,47 @@ Here we provide two Analysis Services:
...
+
+## Getting Started
+### Prerequisites
+#### Kubernetes 
+Crane is compatible with `Kubernetes` version >= 1.18. 
+#### Prometheus
+Crane use prometheus to be the default metric provider. You can install `Prometheus` in or outside `Kubernetes` cluster.
+
+##### Configure Prometheus Address
+The following command will configure your prometheus http address for crane, please change **YOUR_PROMETHEUS_ADDRESS** to actual Prometheus address, for example: `http://prometheus-service.monitoring.svc.cluster.local`
```
mfanjie, line 77: I am thinking that we should have an all-in-one installation so that new comer can install crane without any extra manual config(in a clean env, or we can combine it with kind?).
and we can provide instruction for advance user.

PR #80, opened 2022-01-04
feat: add the feature gate for node resource and cluster node prediction

yan234280533:


pkg/features/features.go
```diff
@@ -9,10 +9,18 @@ import (
...
+	CraneNodeResource featuregate.Feature = "NodeResource"
+
+	// CraneClusterNodePrediction enables the cluster node prediction features.
+	CraneClusterNodePrediction featuregate.Feature = "ClusterNodePrediction"
 )
 
 var defaultFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
-	CraneAutoscaling: {Default: true, PreRelease: featuregate.Alpha},
+	CraneAutoscaling:           {Default: true, PreRelease: featuregate.Alpha},
+	CraneNodeResource:          {Default: false, PreRelease: featuregate.Alpha},
```
qmhu, line 22: How about merge NodeResource and ClusterNodePrediction into one feature gate? If so, we need to use a appropriate name for it.

PR #97, opened 2022-01-11
Add unit tests for match function in analytics controller

alierkilic:
This will address issue: Unit tests for crane #91

If you have any problems with the test names please let me know

PR #101, opened 2022-01-12
Unit test for analytics_controller

wxytjustb:


mfanjie:
@wxytjustb thanks for the PR, but this pr looks duplicated with https://github.com/gocrane/crane/pull/97, there are lots for util functions missing UT, can you take a look and contribute those?

PR #123, opened 2022-02-09
modified the name for node qos ensurance examples

yan234280533:


examples/ensurance/nep-schedule-delay.yaml
```diff
@@ -11,7 +11,7 @@ spec:
     nodeLocalGet:
       localCacheTTLSeconds: 60
   objectiveEnsurances:
-    - name: "container_sched_run_queue_time"
+    - name: "container-schedule-delay-time"
```
mfanjie, line 14: how this metric name map to the actual value?
hard coded in crane-agent?

yan234280533: I will modify it submit a new pr.

PR #127, opened 2022-02-10
Use the same logic to handle completion strategy and add event/condition/logging for analytics 

qmhu:


pkg/controller/recommendation/recommendation_controller.go
```diff
@@ -61,19 +55,48 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+
+	if recommendation.Spec.CompletionStrategy.CompletionStrategyType == analysisv1alph1.CompletionStrategyOnce {
+		if lastUpdateTime != nil {
+			// already finish recommendation
+			return false
+		}
+	} else {
+		if lastUpdateTime != nil {
+			planingTime := lastUpdateTime.Add(time.Duration(*recommendation.Spec.CompletionStrategy.PeriodSeconds) * time.Second)
+			if time.Now().Before(planingTime) {
```
yufeiyu, line 90: Is this too strict here?

PR #130, opened 2022-02-11
refine the node local collector

yan234280533:


pkg/ensurance/collector/cadvisor/cadvisor_linux.go
```diff
@@ -25,69 +24,59 @@ import (
 	"github.com/gocrane/crane/pkg/utils"
 )
 
-const (
-	cadvisorCollectorName = "cadvisor"
-)
-
 func init() {
-	registerMetrics(cadvisorCollectorName, []types.MetricName{types.MetricNameContainerCpuTotalUsage, types.MetricNameContainerSchedRunQueueTime}, NewCadvisor)
+	//registerMetrics(cadvisorCollectorName, []types.MetricName{types.MetricNameContainerCpuTotalUsage, types.MetricNameContainerSchedRunQueueTime}, NewCadvisor)
```
mfanjie, line 28: remove the init() function if it is not needed

pkg/ensurance/collector/cadvisor/cadvisor_linux.go
```diff
@@ -25,69 +24,59 @@ import (
...
+func NewCadvisor(podLister corelisters.PodLister) types.Collector {
 
 	var includedMetrics = cadvisorcontainer.MetricSet{
 		cadvisorcontainer.CpuUsageMetrics:         struct{}{},
 		cadvisorcontainer.ProcessSchedulerMetrics: struct{}{},
 	}
 
-	var allowDynamic bool = true
-	var maxHousekeepingInterval time.Duration = 10 * time.Second
+	var allowDynamic = true
```
mfanjie, line 52: short format `allowDynamic := true` is better when you define a variable with value.

pkg/ensurance/collector/collector.go
```diff
@@ -103,22 +122,60 @@ func (s *StateCollector) UpdateCollectors() {
...
 			s.collectors.Delete(types.NodeLocalCollectorType)
 		}
+
+		// cadvisor need to stop manager before
+		if value, exists := s.collectors.Load(types.CadvisorCollectorType); !exists {
+			s.collectors.Delete(types.CadvisorCollectorType)
+			c := value.(types.Collector)
+			if err = c.Stop(); err != nil {
+				klog.Errorf("Failed to stop the cadvisor manager.")
+				//nothing to do
```
mfanjie, line 158: comment is not needed if the code already shows the intention.

pkg/ensurance/collector/nodelocal/cpu.go
```diff
@@ -1,10 +1,13 @@
 package nodelocal
 
 import (
+	"errors"
 	"fmt"
 	"math"
 	"time"
 
+	"github.com/shirou/gopsutil/load"
```
mfanjie, line 9: org import

pkg/ensurance/collector/nodelocal/cpu.go
```diff
@@ -13,13 +16,13 @@ import (
...
-	cpuCollectorName = "cpu"
-	MAX_PERCENTAGE   = 100
-	MIN_PERCENTAGE   = 0
+	cpuCollectorName     = "cpu"
+	cpuLoadCollectorName = "cpuLoad"
 )
 
 func init() {
-	registerMetrics(cpuCollectorName, []types.MetricName{types.MetricNameCpuTotalUsage, types.MetricNameCpuTotalUtilization}, NewCPUCollector)
+	registerMetrics(cpuCollectorName, []types.MetricName{types.MetricNameCpuTotalUsage, types.MetricNameCpuTotalUtilization}, collectCPU)
```
mfanjie, line 24: registerCollector?
metrics generally means promethus collector

pkg/ensurance/collector/types/interface.go
```diff
@@ -1,11 +1,11 @@
-package collector
+package types
```
mfanjie, line 1: I prefer the previous way, interface should be in the upper folder of implementer, please check k8s code for reference.

pkg/ensurance/collector/types/types.go
```diff
@@ -43,14 +42,24 @@ const (
 )
 
 const (
-	NodeLocalCollectorType CollectType = "node-local"
+	NodeLocalCollectorType     CollectType = "node-local"
```
mfanjie, line 45: types.go has special meaning in k8s, it define the api, maybe const.go is better than types.go?

pkg/ensurance/collector/cadvisor/cadvisor_linux.go
yan234280533: done~

pkg/ensurance/collector/nodelocal/cpu.go
yan234280533: done~

pkg/ensurance/collector/types/types.go
yan234280533: In k8s if it an type define is in types.go. if it only an const value, it in constants.go.

pkg/ensurance/collector/types/interface.go
yan234280533: done~

pkg/ensurance/collector/nodelocal/cpu.go
yan234280533: I modified it.

PR #141, opened 2022-02-21
add validate for NodeQOSEnsurancePolicy and AvoidanceAction

yan234280533:


pkg/ensurance/analyzer/analyzer.go
```diff
@@ -483,7 +483,7 @@ func (s *AnormalyAnalyzer) merge(stateMap map[string][]common.TimeSeries, avoida
 			}
 
 			if action.Spec.Eviction != nil {
-				var deletionGracePeriodSeconds = utils.GetInt32withDefault(action.Spec.Eviction.TerminationGracePeriodSeconds, executor.DefaultDeletionGracePeriodSeconds)
+				var deletionGracePeriodSeconds = utils.GetUint32withDefault(action.Spec.Eviction.TerminationGracePeriodSeconds, executor.DefaultDeletionGracePeriodSeconds)
```
mfanjie, line 486: GetGracePeriodSeconds is better.

pkg/webhooks/ensurance/validating.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	return nil
+}
+
+func validateNodeQualityProbe(nodeProbe ensuranceapi.NodeQualityProbe, fldPath *field.Path) field.ErrorList {
+	var allErrors field.ErrorList
+
+	if (nodeProbe.NodeLocalGet == nil) && (nodeProbe.HTTPGet == nil) {
+		allErrors = append(allErrors, field.Invalid(fldPath.Child("httpGet"), "", "The httpGet and nodeLocalGet must set one"))
```
mfanjie, line 61: The httpGet and nodeLocalGet must set one->
`httpGet and nodeLocalGet cannot be empty at the same time.`

pkg/ensurance/analyzer/analyzer.go
yan234280533: GetInt32withDefault is a generic function for many scenarios such as get piority（utils.GetInt32withDefault(pod.Spec.Priority, 0)）.  

If we rewrite the code like below, it will be too complex.

var deletionGracePeriodSeconds = utils.GetGracePeriodSeconds(action.Spec.Eviction.TerminationGracePeriodSeconds, executor.DefaultDeletionGracePeriodSeconds)

func GetGracePeriodSeconds (i *int32, value int32) int32 {
	return GetInt32withDefault(i,value)
}

func GetInt32withDefault(i *int32, value int32) int32 {
	if i != nil {
		return *i
	}

	return value
}

deploy/manifests/ensurance.crane.io_avoidanceactions.yaml
```diff
@@ -64,8 +66,10 @@ spec:
                       minCPURatio:
                         description: 'MinCPURatio is the min of cpu ratio for low
                           level pods example: the pod limit is 4096, ratio is 10,
-                          the min is 409'
+                          the min is 409 range [0,100]'
```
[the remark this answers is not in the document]
yan234280533: Here is the display question. In my api define is:

	// MinCPURatio is the min of cpu ratio for low level pods
	// example: the pod limit is 4096, ratio is 10, the min is 409
	// range [0,100]

"min is 409" is to explain the example.

I modified it, display like this:

                      minCPURatio:
                        description: 'MinCPURatio is the min of cpu ratio for low
                          level pods. example: the pod limit is 4096, ratio is 10,
                          the minimum is 409. MinCPURatio range [0,100]'
                        format: int64
                        maximum: 100
                        minimum: 0
                        type: integer

api define is:

	// MinCPURatio is the min of cpu ratio for low level pods.
	// example: the pod limit is 4096, ratio is 10, the minimum is 409.
	// MinCPURatio range [0,100]
	// +optional
	// +kubebuilder:validation:Minimum=0
	// +kubebuilder:validation:Maximum=100

deploy/craned/webhooks.yaml
```diff
@@ -26,6 +26,48 @@ webhooks:
...
...MEJBUXNGQURBUU1RNHdEQVlEVlFRRERBVmoKY21GdVpUQWVGdzB5TWpBeU1UY3dPRFEyTlRSYUZ3MHpNakF5TVRVd09EUTJOVFJhTUJBeERqQU1CZ05WQkFNTQpCV055WVc1bE1JSUJJakFOQmdrcWhraUc5dzBCQVFFRkFBT0NBUThBTUlJQkNnS0NBUUVBdm84M25rK1I3eHZXCk15eUE5ZHRTa0RQVUxqcjJSMWVuSFlYaW1JNllDa2pxbWJ3QmQ4enFwNVhPWFNOTEVITW55dWpsaFc0bUJUK2YKUDBsZytoMlVzak5XWTFUQzBUTklyYVNSWmFkd0hlQkhaSEsvM2ZwTlllaDBoTXU3QWFUdmpiYWo4VmZyMGk1TwovL0RHVU5HcjR4VzgwR09PaWxxcDFIcVhaYmM0eGl2ZVBNNFJ5ZmJnbmNQRDdnRytOaDlRYnBjQzBmdmhJVUJHClJFYlBjejFRM2JqTTlQVFEzNTBuZEtwSnY1eGdkaVQ5V1BWZHZVakFSTFhFTDR3TUZaMmVlaUJKUkZLVFhCOG4KSGF0cmJsOU41bHg5ck1ma1NBVFBkMnc4NFFnUUl6M2hSOVlJZXJaS2FJNWtibDN1NjQ3dXZsa015SG1xUHZKSQpwRzlYLzlGd1JRSURBUUFCTUEwR0NTcUdTSWIzRFFFQkN3VUFBNElCQVFCcnlxWXljVmNQOGdMWUVkRU9FR1JQCml1MVNBeFJyTzVXakN3dmNqY2ZueTVVOUJCV1l4VTRPNnl2WTI1ZnVBb1VTNFlweTc5NmxJZWc5UTlyU2EvRDUKTGY1ZVUvTms2YzMzTnhVTjFGdVhtakJSVlNVNkp6YmRPSnFGbVRYYVhnTXRkMGlPUlRUM2xFbWdra0Y2RVdBZAo2dVFrdVJuQkVQVmFhN2xEZk0zV2V5RVRSS2R1dWlOVTEyR1NyQ015RjBiSWZzZ3lyOENDTWtiUzhXWEROeFlhCldPZmpYcEtTYzVJL1ZBL3hRUzJaUEpzK2F1cTMxaEUzOFhjREVFem1RWFdacS95TVJNNVF5UFR2STAwTEtoR2wKaSt1blZMcWFXR1QvVUx2M3JhTFE1eTNlbnNCSUJmYTdIQ2g2LzFPbHJnbHM0aWV1UGd0c1V4Y3ZNVEMrSXk0dAotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0tCg==
```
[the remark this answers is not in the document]
yan234280533: It copied from the  vprediction.crane.io rule. The webhook  server is in the craned not an separate server.

pkg/webhooks/ensurance/validating.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	allNames := sets.String{}
+	for i, obj := range objects {
+		idxPath := fldPath.Index(i)
+
+		if obj.Name == "" {
+			allErrs = append(allErrs, field.Required(idxPath.Child("name"), ""))
+		} else {
+			for _, msg := range validation.IsDNS1123Label(obj.Name) {
```
chenkaiyue, line 114: same with nep name, do we need this to be dns complaint

pkg/webhooks/ensurance/validating.go
```diff
@@ -0,0 +1,253 @@
...
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (p *NepValidationAdmission) ValidateCreate(ctx context.Context, req runtime.Object) error {
+
+	nep, ok := req.(*ensuranceapi.NodeQOSEnsurancePolicy)
+	if !ok {
+		return fmt.Errorf("req can not convert to NodeQOSEnsurancePolicy")
+	}
+
+	allErrs := genericvalidation.ValidateObjectMeta(&nep.ObjectMeta, false, genericvalidation.NameIsDNS1035Label, field.NewPath("metadata"))
```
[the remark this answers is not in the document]
yan234280533: Try to stay consistent with other resources, it restricts all the resource name dns complain.

pkg/webhooks/ensurance/validating.go
yan234280533: It joint an key for the map, like:

var key = strings.Join([]string{n.Name, v.Name}, ".")

so the nep name and object name all used dns complaint.

mfanjie: we should not change user experience, like add constraint to end user because of a implementation challenge.

mfanjie: I am fine after checking with devin together, found many k8s object is using same way to validate name.

PR #148, opened 2022-02-23
Don't predict memory in effective hpa

qmhu:


pkg/recommend/advisor/ehpa.go
```diff
@@ -96,11 +96,11 @@ func (a *EHPAAdvisor) Advise(proposed *types.ProposedRecommendation) error {
 	targetCpuUtilization := int32(50) // todo: configurable
 	maxReplicasFactor := 1.2          // todo: configurable
 
-	maxCpuUtilization := int32(int64(maxCpuUsage) * 1000 * 100 / requestTotal)
+	maxCpuUtilization := int32(int64(maxCpuUsage) * 1000 * 100 * int64(len(a.Pods)) / requestTotal)
```
kitianFresh, line 99: why multiply pods num. maxCpuUsage is average max usage of each pod?  I see it is computed only by one timeseries tsList[0], not all

pkg/recommend/advisor/ehpa.go
```diff
@@ -96,11 +96,11 @@ func (a *EHPAAdvisor) Advise(proposed *types.ProposedRecommendation) error {
...
 	proposedMaxRatio := float64(maxCpuUtilization) / float64(targetCpuUtilization)
 	maxReplicasProposed := int32(math.Ceil(proposedMaxRatio * float64(a.ReadyPodNumber) * maxReplicasFactor))
 
-	minCpuUtilization := int32(int64(minCpuUsage) * 1000 * 100 / requestTotal)
+	minCpuUtilization := int32(int64(minCpuUsage) * 1000 * 100 * int64(len(a.Pods)) / requestTotal)
```
kitianFresh, line 103: same fuzzing as above

pkg/controller/ehpa/hpa.go
```diff
@@ -189,7 +189,7 @@ func (c *EffectiveHPAController) GetHPAMetrics(ctx context.Context, ehpa *autosc
 			if metric.Type == autoscalingv2.ResourceMetricSourceType {
 				name, err := GetPredictionMetricName(metric.Resource.Name)
 				if err != nil {
-					return nil, err
+					continue
```
mfanjie, line 192: log error, suggest not make it silent

pkg/controller/ehpa/predict.go
```diff
@@ -131,7 +131,7 @@ func (c *EffectiveHPAController) NewPredictionObject(ehpa *autoscalingapi.Effect
 		if metric.Type == autoscalingv2.ResourceMetricSourceType {
 			metricName, err := GetPredictionMetricName(metric.Resource.Name)
 			if err != nil {
-				return nil, err
+				continue
```
mfanjie, line 134: same as above

pkg/recommend/advisor/ehpa.go
qmhu: requestTotal contains total request for workload, so we need to add this.

kitianFresh: maxCpuUsage is max of workload‘s(sum of all pods) usage too, so the original code is right, do not need fix

PR #154, opened 2022-02-25
[WIP, DO NOT MERGE] refactor analyze

chenkaiyue:


examples/ensurance/waterline1.yaml
```diff
@@ -11,11 +11,11 @@ spec:
     nodeLocalGet:
```
yan234280533, line 11: merge the commit to the only one

PR #168, opened 2022-02-28
add get cluster's namespace interface; use nginx to connect frontend

chenkaiyue:


deploy/craned/deployment.yaml
```diff
@@ -35,7 +35,7 @@ spec:
           operator: Exists
       containers:
         - name: craned
-          image: docker.io/gocrane/craned:v0.2.0
+          image: ccr.ccs.tencentyun.com/chenky/crane:022101
```
kitianFresh, line 38: Do not modify the image.

deploy/craned/deployment.yaml
```diff
@@ -49,6 +49,17 @@ spec:
               readOnly: true
             - name: config
               mountPath: /tmp/recommendation-config
+          readinessProbe:
+            httpGet:
+              path: /api/healthz
+              port: 8082
+              scheme: HTTP
+        - image: gcr.io/kubecost1/frontend:prod-1.90.0-rc.0
```
kitianFresh, line 57: Please replace this image

deploy/craned/service.yaml
```diff
@@ -11,3 +11,31 @@ spec:
...
+  namespace: crane-system
+spec:
+  ports:
+    - port: 8082
+      protocol: TCP
+      targetPort: 8082
+  selector:
+    app: craned
+---
+apiVersion: v1
```
kitianFresh, line 28: I think we should only expose one service name crane. the service will expose multiple ports, include crane(which is the crane nginx web endpoint), craned-http (which is craned controller), craned web-http (which is craned web server), craned metrics-http (which is craned metrics)  @qmhu

deploy/craned/deployment.yaml
chenkaiyue: Ok

chenkaiyue: Ok

deploy/craned/deployment.yaml
chenkaiyue: Just for test

chenkaiyue: Just for test

deploy/craned/service.yaml
chenkaiyue: Ok

deploy/craned/deployment.yaml
```diff
@@ -49,6 +49,20 @@ spec:
...
+              port: 8082
+              scheme: HTTP
+        - image: docker.io/gocrane/dashboard:v0.2.0
+          imagePullPolicy: IfNotPresent
+          name: dashboard
+          volumeMounts:
+            - mountPath: /etc/nginx/conf.d/
+              name: nginx-conf
+          env:
+          - name: CLUSTER_ID
```
kitianFresh, line 64: What is this used for? the front end use ?

PR #178, opened 2022-03-02
Default PROMQL query syntax should not use regular expressions

shijieqin:
NodeMemUsagePromQLFmtStr = `sum(node_memory_MemTotal_bytes{instance="%s"} - node_memory_MemAvailable_bytes{instance="%s"})`
NodeCpuUsagePromQLFmtStr = `sum(count(node_cpu_seconds_total{mode="idle",instance="%s"}) by (mode, cpu)) - sum(irate(node_cpu_seconds_total{mode="idle",instance="%s"}[%s]))`
The above two default queries use a regular expression, which will cause the query result that does not meet the expected expectations.

### DIFF
--- a/pkg/prediction/config/config.go
+++ b/pkg/prediction/config/config.go
@@ -19,9 +19,9 @@ const (
 
 	// following is node exporter metric for node cpu/memory usage
 	// NodeCpuUsagePromQLFmtStr is used to query node cpu usage by promql,  param is node name which prometheus scrape, duration str
-	NodeCpuUsagePromQLFmtStr = `sum(count(node_cpu_seconds_total{mode="idle",instance=~"%s.*"}) by (mode, cpu)) - sum(irate(node_cpu_seconds_total{mode="idle",instance=~"%s.*"}[%s]))`
+	NodeCpuUsagePromQLFmtStr = `sum(count(node_cpu_seconds_total{mode="idle",instance=~"(%s)(:\\d+)?"}) by (mode, cpu)) - sum(irate(node_cpu_seconds_total{mode="idle",instance=~"(%s)(:\\d+)?"}[%s]))`
 	// NodeMemUsagePromQLFmtStr is used to query node cpu memory by promql,  param is node name, node name which prometheus scrape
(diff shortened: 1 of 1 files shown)


pkg/prediction/config/config.go
```diff
@@ -19,9 +19,9 @@ const (
 
 	// following is node exporter metric for node cpu/memory usage
 	// NodeCpuUsagePromQLFmtStr is used to query node cpu usage by promql,  param is node name which prometheus scrape, duration str
-	NodeCpuUsagePromQLFmtStr = `sum(count(node_cpu_seconds_total{mode="idle",instance=~"%s.*"}) by (mode, cpu)) - sum(irate(node_cpu_seconds_total{mode="idle",instance=~"%s.*"}[%s]))`
+	NodeCpuUsagePromQLFmtStr = `sum(count(node_cpu_seconds_total{mode="idle",instance="%s"}) by (mode, cpu)) - sum(irate(node_cpu_seconds_total{mode="idle",instance="%s"}[%s]))`
 	// NodeMemUsagePromQLFmtStr is used to query node cpu memory by promql,  param is node name, node name which prometheus scrape
-	NodeMemUsagePromQLFmtStr = `sum(node_memory_MemTotal_bytes{instance=~"^%s.*"} - node_memory_MemAvailable_bytes{instance=~"^%s.*"})`
+	NodeMemUsagePromQLFmtStr = `sum(node_memory_MemTotal_bytes{instance="%s"} - node_memory_MemAvailable_bytes{instance="%s"})`
```
kitianFresh: This regex is used to match some instance including ip:port style. And it depends on your monitoring system, so we do not assume the instance is nodename or ip only.  

If we modify this prom template, we need find a more general way.

image

shijieqin: I know, so I pushed a new commit that is used to match some instance including ip:port style.
image

mfanjie:
@shijieqin cloud you please make sure your commit messages are in English, we need to keep the project to be globalized.

shijieqin:
> @shijieqin cloud you please make sure your commit messages are in English, we need to keep the project to be globalized.

Sorry, I change the commit message.

pkg/prediction/config/config.go
kitianFresh: \LGTM

PR #187, opened 2022-03-07
add makefile for crane dashboard

chenkaiyue:
#### What type of PR is this?
Add crane dashboard in makefile.


Makefile
```diff
@@ -110,7 +111,7 @@ test: fmt vet lint ## Run tests.
 build: craned crane-agent metric-adapter
 
 .PHONY: all
-all: generate test craned  crane-agent metric-adapter
+all: generate test craned  crane-agent metric-adapter image-dashboard
```
mfanjie, line 114: shouldn't be dashboard like others?
make all are not building images.

chenkaiyue: My fault, i put it into images.

PR #191, opened 2022-03-08
break loop once cronScaler found

yufeiyu:
Break loop once cronScaler found

Fixes #


pkg/metricprovider/external_metric_provider.go
```diff
@@ -49,11 +49,13 @@ func (p *ExternalMetricProvider) GetExternalMetric(ctx context.Context, namespac
 	}
 	// Find the cron metric scaler
 	var cronScaler *CronScaler
+outer:
```
kitianFresh, line 52: We can use a function to do this, if find return the cron.

yufeiyu: done

PR #193, opened 2022-03-10
modify some format and comment for crane-agent

yan234280533:
#### What type of PR is this?
format and comment modified

modify some format and comment for crane-agent

Fixes #

### DIFF
--- a/pkg/metrics/ensuarance.go
+++ b/pkg/metrics/ensuarance.go
@@ -63,7 +63,8 @@ const (
 )
 
 var (
-	/**** Metrics related to crane-agent execution ****/
+
+	// LastActivity records the last activity time of each steps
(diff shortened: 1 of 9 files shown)


pkg/metrics/ensuarance.go
```diff
@@ -63,7 +63,8 @@ const (
 )
 
 var (
-	/**** Metrics related to crane-agent execution ****/
+
+	// LastActivity used to record the last activity time of each steps
```
mfanjie, line 67: `// LastActivity records the last activity time of each steps`
remove `used to`

yan234280533: done~

mfanjie:
lgtm but need rebase @yan234280533

PR #195, opened 2022-03-10
[WIP] offline manager

chenkaiyue:
#### What type of PR is this?
Feature

1. move node resource controller as offline manager to agent
2. add offline job qos ensurance feature in offline manager
3. add a data source, offline manager can get data from both tsp and local


mfanjie:
`move node resource controller as offline manager to agent` I think @szy441687879 was about to do the same thing? can you guys have a sync?

shijieqin:
> `move node resource controller to agent` I think @szy441687879 was about to do the same thing? can you guys have a sync?

we are syncing

PR #196, opened 2022-03-11
Convert Pod ext resource to quota limit

chenkaiyue:
#### What this PR does / why we need it:
1. move node resource controller as node resource manager to agent
2. add low priority pod which used ext resource qos ensurance feature in node resource manager
3. add a data source, node resource manager can get data from both tsp and local

PR #165, opened 2022-02-28
add advanced cpuset manger

szy441687879:
# Advanced CPUSet Manager

- Provide a new mechanism to manage cpuset bypass, to do extra things that kubelet cannot do, such as revise cpuset when pod running, share the cpuset assigned with others and relax restrictions of cpuset allocation.

## Table of Contents

- [Advanced CPUSet Manager](#advanced-cpuset-manager)
    - [Table of Contents](#table-of-contents)
    - [Motivation](#motivation)
        - [Goals](#goals)
        - [Non-Goals/Future Work](#non-goalsfuture-work)
    - [Proposal](#proposal)
        - [Relax restrictions of cpuset allocation](#relax-restrictions-of-cpuset-allocation)
        - [Add new annotation to describe the  requirement of cpuset contorl manger](#add-new-annotation-to-describe-the--requirement-of-cpuset-contorl-manger)
        - [Advanced CPU Manager component](#advanced-cpu-manager-component)
        - [User Stories](#user-stories)
            - [Story 1](#story-1)
            - [Story 2](#story-2)
        - [Risks and Mitigations](#risks-and-mitigations)

## Motivation
Some latency-sensitive application have lower lantency and cpu usage when running with specific cores, which results in fewer context switchs and higer cache affinity.
But kubelet will always exclude assigned cores in shared cores, which may waste resources.Offline and other online pods can running on the codes actually. In our experiment, for the most part, it is barely noticeable for performance of service.

### Goals

- Provide a new mechanism to manage cpuset bypass
- Provide a new cpuset control way "shared"
- Allow revise cpuset when pod running
- Relax restrictions of cpuset allocation

### Non-Goals/Future Work

- Solve the conflict with kubelet static cpuset control (need close)
- Numa manager will support in future, CCX/CCD manager can also considered

### Relax restrictions of cpuset allocation
Kubelet allocate cpus for containers included in a pod of "Guaranteed" QoS class and the container's resource limit for the CPU resource is an integer greater than or equal to one.

With Crane, containers only to neeed CPU resource limit is an integer greater than or equal to one and equal to cpu resource request.

### Add new annotation to describe the  requirement of cpuset contorl manger
```yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    qos.gocrane.io/cpu-manager: none/exclusive/share
```
Provide three policy for cpuset control:
- none: containers of this pod shares a set of cpus not allocated top exclusive policy containers
[...]

PR #197, opened 2022-03-11
exporter CPU metrics f ext-servervice and CPU metrics of nodes that can be reused

shijieqin:
exporter CPU metrics f ext-servervice and CPU metrics of nodes that can be reused

exporter CPU metrics f ext-servervice 
exporter CPU metrics of nodes that can be reused


pkg/metrics/resource_metrics_collector.go
```diff
@@ -0,0 +1,92 @@
...
+		cpuIdleCanBeReused = rc.cpuStateProvider.CollectCpuCoresCanBeReused()
+	}()
+	go func() {
+		defer wg.Done()
+		offlineCpuUsageIncrease, offlineCpuUsageAvg = rc.cpuStateProvider.GetExtCpuUsage()
+	}()
+	lastTime := time.Now()
+	wg.Wait()
+
+	atomic.AddUint64(&cpuUsageCounter, offlineCpuUsageIncrease)
```
chenkaiyue, line 78: cpuUsageCounter use offlineCpuUsageIncrease, while according to `getExtCpuUsageFromK8s`, for a new pod, the diff will only accumulate the diff between next collect, the init value when called `getExtCpuUsageFromK8s` is lost. Is this designed on purpose?

pkg/agent/agent.go
```diff
@@ -57,6 +75,9 @@ func NewAgent(ctx context.Context,
 	managers = append(managers, analyzerManager)
 	avoidanceManager := executor.NewActionExecutor(kubeClient, nodeName, podInformer, nodeInformer, noticeCh, runtimeEndpoint)
 	managers = append(managers, avoidanceManager)
+	cpuManager := cm.NewAdvancedCpuManager(kubeClient, nodeName, podInformer, nodeInformer, runtimeEndpoint, cadvisorManager)
+	managers = append(managers, cpuManager)
+	metrics.RegisterResourceMetrics(agent.host, utils.NewCpuStateProvider(cadvisorManager, podInformer.Lister(), useBt, cpuManager.GetExclusiveCpu))
```
chenkaiyue, line 80: After CpuStateProvider merge into collector, we could use data from collecter here.

chenkaiyue:
A question to discuss:
In the current design, the metric is calculated and exposed every time prometheus pulls. The advantage is the data is newest, the disadvantage is we want cpuStateProvider merge into collector, this method `CollectCpuCoresCanBeReused` may should be exposed, at the beginning, we just want to use data from collector. And this will increased agent's computational pressure.

If we use calculated data like other metrics, the point will not accurate. For example, prometheus pull period is 10s, the collector collect period is 25s, the point promethues will not be precise and may lost some data point from collector's data.

pkg/agent/agent.go
shijieqin: Yes~

pkg/utils/cadvisor_linux.go
```diff
@@ -0,0 +1,236 @@
+package utils
+
```
yan234280533, line 2: Can you do a rebase of main, we will continue the pr.

pkg/utils/bt/cpu.go
```diff
@@ -0,0 +1,196 @@
...
+
+//copy from "github.com/shirou/gopsutil/cpu"
+// Total returns the total number of seconds in a CPUTimesStat
+func (c TimesStat) Total() float64 {
+	total := c.User + c.System + c.Nice + c.Iowait + c.Irq + c.Softirq +
+		c.Steal + c.Idle
+	return total
+}
+
+func (c TimesStat) CalculateBtOffline(stat2 TimesStat) (float64, float64) {
```
yan234280533, line 180: not use bt and offline, use priority instead

pkg/utils/string.go
```diff
@@ -17,3 +20,15 @@ func ParseFloat(str string, defaultValue float64) (float64, error) {
 	}
 	return strconv.ParseFloat(str, 64)
 }
+
+// parsePercentage parse the percent string value
+func ParsePercentage(input string) (float64, error) {
```
yan234280533, line 25: Can we use a more generic way to define the function ？

yan234280533:
@shijieqin As much code is merged to the main branch in pr #165 and #196, please do a rebase of the main branch, we will continue the review of this pr after this.

pkg/metrics/resource_metrics_collector.go
```diff
@@ -0,0 +1,87 @@
...
+	"k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
+)
+
+var (
+	nodeExtCPUUsageDesc = prometheus.NewDesc("node_ext_cpu_usage_seconds_total",
+		"Cumulative cpu time consumed by the node in core-seconds",
+		[]string{"node"},
+		nil)
+	nodeCPUCanBeReusedDesc = prometheus.NewDesc("node_cpu_can_be_reused_seconds",
```
yan234280533, line 19: node_cpu_can_be_reused_seconds
-> 
node_cpu_reallocated

pkg/metrics/resource_metrics_collector.go
```diff
@@ -0,0 +1,87 @@
...
+		"Cumulative cpu time consumed by the node in core-seconds",
```
yan234280533, line 20: Cumulative cpu time consumed by the node in core-seconds
->
Reallocated CPU consumed by the node in cores

shijieqin:
https://github.com/gocrane/crane/pull/223

PR #198, opened 2022-03-11
add NodeResourceManager

shijieqin:
1. Refactoring ext resource logic in crane-agent.
2. Merge real-time data of kubernetes nodes to assist NodeResourceManager in calculating ext resource.
3. Creates a tsp for predicting reusable resources when the crane-agent starts.


pkg/agent/agent.go
```diff
@@ -95,5 +128,64 @@ func (a *Agent) Run(healthCheck *metrics.HealthCheck, opts *options.Options) {
...
+		}
+	}
+	config, err := a.kubeClient.CoreV1().ConfigMaps("default").Get(context.TODO(), "noderesource-tsp-template", metav1.GetOptions{})
+
+	if err != nil {
+		klog.Exitf("Get noderesource tsp configmap noderesource-tsp-template with error: %v", err)
+	}
+
+	if config == nil {
+		klog.Exitf("Can't get noderesource tsp configmap noderesource-tsp-template")
```
chenkaiyue, line 151: I found many "exitf"? Do we need exitf here?

pkg/agent/agent.go
```diff
@@ -95,5 +128,64 @@ func (a *Agent) Run(healthCheck *metrics.HealthCheck, opts *options.Options) {
...
+	}
+
+	spec := v1alpha12.TimeSeriesPredictionSpec{}
+	err = yaml.Unmarshal([]byte(strings.Replace(config.Data["spec"], "{{nodename}}", a.host, -1)), &spec)
+	if err != nil {
+		klog.Exitf("Convert spec template error: %v", err)
+	}
+
+	n, _ := a.kubeClient.CoreV1().Nodes().Get(context.TODO(), a.host, metav1.GetOptions{})
```
chenkaiyue, line 160: Error should be handled

pkg/utils/cadvisor_linux.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+func NewCadvisorProvider(manager cmanager.Manager, podLister corelisters.PodLister) *CadvisorProvider {
+	return &CadvisorProvider{
+		Manager:   manager,
+		podLister: podLister,
+	}
+}
+
+func (c *CadvisorProvider) GetCPUAndMemoryStats() (*statsapi.Summary, error) {
```
chenkaiyue, line 65: I think all these logic should be merged into collector module, will be more in line with the design principles of each module.

pkg/utils/cadvisor_linux.go
```diff
@@ -0,0 +1,236 @@
...
+		result = append(result, *podStats)
+	}
+	return &statsapi.Summary{
+		Pods: result,
+	}, nil
+}
+
+// cadvisorInfoToContainerCPUAndMemoryStats returns the statsapi.ContainerStats converted
+// from the container and filesystem info.
+func cadvisorInfoToContainerCPUAndMemoryStats(name string, info *cadvisorapiv2.ContainerInfo) *statsapi.ContainerStats {
```
chenkaiyue, line 122: reference mark

pkg/utils/bt/cpu.go
```diff
@@ -0,0 +1,196 @@
...
+	GuestNice float64 `json:"guestNice"`
+	Offline   float64 `json:"offline"`
+}
+
+func Times(percpu bool) ([]TimesStat, error) {
+	return TimesWithContext(context.Background(), percpu)
+}
+
+func TimesWithContext(ctx context.Context, percpu bool) ([]TimesStat, error) {
+	filename := HostProc("bt_stat")
```
chenkaiyue, line 38: Bt is better not to appear in the project. In crane we don't rely on it.

pkg/utils/cpu.go
```diff
@@ -0,0 +1,256 @@
...
+
+type CpuStateProvider struct {
+	podLister       corelisters.PodLister
+	provider        *CadvisorProvider
+	mu              sync.RWMutex
+	extPodState     map[summary.PodReference]summary.PodStats
+	cpuState        *CpuTimeStampState
+	btCpuState      *BtCpuTimeStampState
+	exclusiveCPUSet func() cpuset.CPUSet
+	useBt           bool
```
chenkaiyue, line 63: Bt is better not to appear in the project. In crane we don't rely on it.

pkg/utils/cpu.go
```diff
@@ -0,0 +1,256 @@
...
+		klog.V(4).Infof("resourceMetricsCollector bt cpu state init")
+		return 0, 0
+	}
+	offlineIncrease, offlinePercent := rc.btCpuState.stat.CalculateBtOffline(stat)
+	rc.btCpuState = nowCpuState
+
+	return uint64(offlineIncrease * float64(time.Second)), offlinePercent * float64(rc.cpuCoreNumbers) / 100
+}
+
+func (rc *CpuStateProvider) getExtCpuUsageFromK8s() (uint64, float64) {
```
chenkaiyue, line 102: I think it's better to put this logic and extPodState  into collector

pkg/utils/bt/common.go
```diff
@@ -0,0 +1,69 @@
+package bt
```
chenkaiyue, line 1: Bt is better not to appear in the project. In crane we don't rely on it.

pkg/noderesource/noderesource.go
```diff
@@ -0,0 +1,340 @@
...
+		}
+	}()
+}
+
+func (nr *NodeResource) GenerateFullyAnalyzeData(state map[string][]MetricTimeSeries, collectorName string) map[string][]MetricTimeSeries {
+	for _, c := range nr.collectorList {
+		if c.Name() == collectorName {
+			continue
+		}
+		tmpState := c.GetLastState()
```
chenkaiyue, line 165: Make tsp and local value together, and find the smallest one?

pkg/noderesource/noderesource.go
```diff
@@ -0,0 +1,340 @@
...
+	//获取指标上限值
+	for resourceName, mTsList := range state {
+		for _, metricTimeSeries := range mTsList {
+			if metricTimeSeries.DataSourceName != GetRealtimeCollectorName() {
+				continue
+			}
+			for _, timeSeries := range metricTimeSeries.TimeSeriesList {
+				for _, sample := range timeSeries.Samples {
+					if lowerLimitResourceNameMap[resourceName] < sample.Value*0.9 {
+						lowerLimitResourceNameMap[resourceName] = sample.Value * 0.9
```
chenkaiyue, line 196: What is 0.9 mean? Maybe configurable?

pkg/noderesource/noderesource.go
```diff
@@ -0,0 +1,340 @@
...
+	klog.V(4).Infof("state of NodeResource: %v", state)
+	node, err := nr.nodeLister.Get(nr.nodeName)
+	if err != nil {
+		klog.Errorf("Failed to get node: %v", err)
+		return
+	}
+	nodeCopy := node.DeepCopy()
+	effectDataSourceNameMap := make(map[string]string)
+	lowerLimitResourceNameMap := make(map[string]float64)
+	//获取指标上限值
```
chenkaiyue, line 187: Use english comment

pkg/noderesource/noderesource.go
```diff
@@ -0,0 +1,340 @@
...
+		oldJson, err = json.Marshal(node)
+		if err == nil {
+			newJson, err = json.Marshal(nodeCopy)
+			if err == nil {
+				jsonPatch, err = strategicpatch.CreateTwoWayMergePatch(oldJson, newJson, nodeCopy)
+				klog.V(4).Infof("jsonPatch: %s", jsonPatch)
+			}
+		}
+
+		if err != nil {
```
chenkaiyue, line 247: The logic is to use jsonPatch with PatchStatus first, if err,  use UpdateStatus?
Why not use UpdateStatus directly?

chenkaiyue:
This PR calculate the idle resource, so tsp prefer larger value, the predict result will affect online business. I discussed with @kitianFresh , we can only adjust parameters, so it's better to get and expose a reverse value: all-(cpuIdleCanBeReused + offlineCpuUsageAvg)

chenkaiyue:
I suggest move cpu and cadvisor, include realtime in noderesource into collector; tsp.go can put into noderesource and still watch tsp; bt related should be removed.

pkg/agent/agent.go
shijieqin: After nodesource is enabled, if I can't get the template of tsp, I want the program to quit instead of continuing to run.

pkg/agent/agent.go
shijieqin: Eh... I just think the crane-agent is up, so it should be possible to get nodes normally. Ha, I'll change the code.

pkg/utils/cadvisor_linux.go
chenkaiyue: This func is no need? 
cpuUsageSample, schedRunqueueTime := caculateCPUUsage(&v, &state) in cadvisor_linux.go calculate the cpu usage, which is just the usage diff/time diff, we can directly use it. The name container_cpu_total_usage is really misleading.

chenkaiyue:
The calculated cadvisor related funcs are no need?
cpuUsageSample, schedRunqueueTime := caculateCPUUsage(&v, &state) in cadvisor_linux.go calculate the cpu usage, which is just the usage diff/time diff, we can directly use it. The name container_cpu_total_usage is really misleading.

I think you can expand base on containerStates in cadvisor_linux.go.

shijieqin:
https://github.com/gocrane/crane/pull/219

PR #205, opened 2022-03-17
make cadvisor a member variable in collector

chenkaiyue:
#### What type of PR is this?
feature

We found that many modules need to use cadvisor in collector, such as pod resource manager, cpu set manager, while only when nep has nodelocal configure cadvisor can exist.

So make cadvisor a  member variable in collector and be configuration independent.

### DIFF
--- a/pkg/ensurance/collector/cadvisor/cadvisor_linux.go
+++ b/pkg/ensurance/collector/cadvisor/cadvisor_linux.go
@@ -45,8 +45,7 @@ type CadvisorCollector struct {
 	latestContainersStates map[string]ContainerState
 }
 
-func NewCadvisor(podLister corelisters.PodLister) *CadvisorCollector {
-
+func NewCadvisorManager() cmanager.Manager {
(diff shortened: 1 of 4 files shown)


pkg/ensurance/collector/cadvisor/cadvisor_linux.go
```diff
@@ -45,8 +45,7 @@ type CadvisorCollector struct {
 	latestContainersStates map[string]ContainerState
 }
 
-func NewCadvisor(podLister corelisters.PodLister) *CadvisorCollector {
-
+func NewCadvisorManager() cmanager.Manager {
```
szy441687879, line 48: "github.com/google/cadvisor/manager" can not build succfullly without linux,
so cmanager.Manager can not be the output parameters of function.
you can refer to。
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cadvisor/cadvisor_unsupported.go

PR #223, opened 2022-03-22
【WIP】Export cpu metrics that can't be reclaimed

shijieqin:
1. Export cpu metrics that can't be reclaimed


mfanjie:
fixed in other PR

PR #239, opened 2022-03-30
refine querybuilder factory to register way, to avoid switch case mod…

### LINKED ISSUE #238: refine querybuilder factory to register way, to avoid switch case mode
## Describe the feature
Now the querybuilder factory use hard code switch case to select different type builder, this is not a code friendly way in Golang.

Refine it to map factory, and register each implementation by Go init func in each implementer.

kitianFresh:
…e. each implementer only register by init way

#### What type of PR is this?
refine querybuilder factory to register way

Fixes #238

### DIFF
--- a/pkg/querybuilder/query.go
+++ b/pkg/querybuilder/query.go
@@ -1,6 +1,8 @@
 package querybuilder
 
 import (
+	"sync"
+
 	"github.com/gocrane/crane/pkg/metricquery"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/querybuilder/query.go
```diff
@@ -13,3 +15,22 @@ type Builder interface {
...
 }
+
+var (
+	factoryLock    sync.Mutex
+	builderFactory = make(map[metricquery.MetricSource]BuilderFactoryFunc)
+)
+
+type BuilderFactoryFunc func(metric *metricquery.Metric) Builder
+
+func RegisterBuilderFactory(metricSource metricquery.MetricSource, initFunc BuilderFactoryFunc) {
```
qmhu, line 26: Do we need lock here?

PR #258, opened 2022-04-11
refine cron metrics priority for ehpa when working with other hpa met…

### LINKED ISSUE #257: refine cron metric priority when working with other metric together
## Describe the feature
Now when cron for ehpa works with no other metrics, it can works well.  But when there are other metrics such as prediction or original hpa metrics, it has some confilicts with other metrics when it is not in active.

Refine the cron pirority to ignore it when it is not active while working with other metrics together

kitianFresh:
refine cron metrics priority for ehpa when working with other hpa metrics together, remove the cron impact when it is not active

#### What type of PR is this?
Feature

Fixes #257

### DIFF
--- a/pkg/controller/ehpa/effective_hpa_controller.go
+++ b/pkg/controller/ehpa/effective_hpa_controller.go
@@ -76,7 +76,7 @@ func (c *EffectiveHPAController) Reconcile(ctx context.Context, req ctrl.Request
 	}
 
 	// reconcile prediction if enabled
-	if IsPredictionEnabled(ehpa) {
+	if utils.IsEHPAPredictionEnabled(ehpa) {
 		prediction, err := c.ReconcilePredication(ctx, ehpa)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/utils/ehpa/ehpa.go
```diff
@@ -0,0 +1,11 @@
+package ehpa
+
```
qmhu, line 2: I'd rather not create folder for ehpa.go in utils.

kitianFresh: > I'd rather not create folder for ehpa.go in utils.

Then it needs distinguish it is EHPA util function, so function name need to be IsEHPAPredictionEnabled and IsEHPACronEnabled

qmhu: these two function is only available for ehpa, so your origin name can show the meaning for the function . However IsEHPAPredictionEnabled and IsEHPACronEnabled is fine to me

PR #278, opened 2022-04-22
support init mode for metric model

### LINKED ISSUE #212: Make it is optional when init the percentile histogram model
## Describe the feature
Now percentile algorithm prediction is sync and block when init from history data source provider. we make it optional by three ways:
1. init from history data source provider directly just as now.
2. init from a checkpoint directly, but this checkpoint storage should be defined, crd, or remote storage
3. lazy init and accumulating a period of time by real time data source provider metrics, when it is ready, then it can do predict.

kitianFresh:
…async until the data is accumlulating enough

#### What type of PR is this?
Feature

support init percentile model by different modes, by default, it always use history mode, keep the original logic.
for recommendation, it is once task use original algorithm, so it does not impact.
for tsp, it has params and caller, so it does not impact.
for evpa, use lazytrainning mode to do predict until it is accumulating data to enough window. this way the prediction is more robust and safe.

Fixes #212

### DIFF
--- a/pkg/autoscaling/estimator/percentile.go
+++ b/pkg/autoscaling/estimator/percentile.go
@@ -19,7 +19,7 @@ import (
 	"github.com/gocrane/crane/pkg/utils"
 )
 
-const callerFormat = "EVPACaller-%s-%s-%s"
+const callerFormat = "EVPACaller-%s-%s"
 
 type PercentileResourceEstimator struct {
 	Predictor prediction.Interface
@@ -29,7 +29,9 @@ type PercentileResourceEstimator struct {
 func (e *PercentileResourceEstimator) GetResourceEstimation(evpa *autoscalingapi.EffectiveVerticalPodAutoscaler, config map[string]string, containerName string, currRes *corev1.ResourceRequirements) (corev1.ResourceList, error) {
 	recommendResource := corev1.ResourceList{}
 
+	caller := fmt.Sprintf(callerFormat, klog.KObj(evpa), string(evpa.UID))
 	cpuMetricNamer := &metricnaming.GeneralMetricNamer{
+		CallerName: caller,
 		Metric: &metricquery.Metric{
 			Type:       metricquery.ContainerMetricType,
 			MetricName: corev1.ResourceCPU.String(),
@@ -43,7 +45,7 @@ func (e *PercentileResourceEstimator) GetResourceEstimation(evpa *autoscalingapi
 	}
 
 	cpuConfig := getCpuConfig(config)
-	tsList, err := utils.QueryPredictedValues(e.Predictor, fmt.Sprintf(callerFormat, string(evpa.UID), containerName, corev1.ResourceCPU), cpuConfig, cpuMetricNamer)
+	tsList, err := utils.QueryPredictedValues(e.Predictor, caller, cpuConfig, cpuMetricNamer)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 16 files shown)


qmhu:
/LGTM 
@yufeiyu please take a look

pkg/autoscaling/estimator/percentile.go
```diff
@@ -137,9 +142,22 @@ func getCpuConfig(config map[string]string) *predictionconfig.Config {
 		marginFraction = "0.15"
 	}
 
+	initModeStr, exists := config["cpu-model-init-mode"]
+	initMode := predictionconfig.ModelInitModeLazyTraining
+	if !exists {
+		initMode = predictionconfig.ModelInitMode(initModeStr)
+	}
+
+	historyLength, exists := config["cpu-model-history-length"]
```
xieydd, line 151: How about moving the constant to a common file, like `EVPACaller-%s-%s` 、 `cpu-model-history-length`、`24h` etc.?

pkg/autoscaling/estimator/percentile.go
```diff
@@ -166,9 +184,22 @@ func getMemConfig(props map[string]string) *predictionconfig.Config {
 		marginFraction = "0.15"
 	}
 
+	initModeStr, exists := props["mem-model-init-mode"]
```
xieydd, line 187: This part of memory is almost same like cpu, how about a function ?

pkg/autoscaling/estimator/percentile.go
```diff
@@ -166,9 +184,22 @@ func getMemConfig(props map[string]string) *predictionconfig.Config {
...
+	initMode := predictionconfig.ModelInitModeLazyTraining
+	if !exists {
+		initMode = predictionconfig.ModelInitMode(initModeStr)
+	}
+
+	historyLength, exists := props["mem-model-history-length"]
+	if !exists {
```
xieydd, line 194: Why history length of cpu and memory is not same, cpu is 24h but memory is 48h?

pkg/prediction/config/types.go
```diff
@@ -10,7 +10,20 @@ type AlgorithmModelConfig struct {
 	UpdateInterval time.Duration
 }
 
+type ModelInitMode string
+
+const (
+	// means recover or init the algorithm model directly from history datasource, this process may block because it is time consuming for data fetching & model gen
```
xieydd, line 16: Can we make sure which  mode is default?

pkg/prediction/percentile/prediction.go
```diff
@@ -276,14 +286,31 @@ func (p *percentilePrediction) Run(stopCh <-chan struct{}) {
...
-			if err := p.init(qc.MetricNamer); err != nil {
-				klog.ErrorS(err, "Failed to init percentilePrediction.")
+			c := p.a.GetConfig(QueryExpr)
+			if c == nil {
+				c = &defaultInternalConfig
+			}
+
+			var initError error
+			switch c.initMode {
+			case config.ModelInitModeLazyTraining:
```
xieydd, line 296: lazy training mode is not from history?

xieydd (changes requested):
Some nits.

pkg/prediction/config/types.go
kitianFresh: this param is specified by caller or user, if it is not specified, default is history mode, original logic

pkg/prediction/percentile/prediction.go
kitianFresh: > lazy training mode is not from history?

yes, it means accumulating data from real time data source, then when the data length is enough, it can do predict. i have not thought a better naming

pkg/autoscaling/estimator/percentile.go
kitianFresh: > Why history length of cpu and memory is not same, cpu is 24h but memory is 48h?

This is just an empirical value borrowed from vpa.  memory is incompressible resource, so use longer history data is more safe and robust. cpu is compressible resource, and generally it is daily cycle because of people traffic

pkg/autoscaling/estimator/percentile.go
kitianFresh: > How about moving the constant to a common file, like `EVPACaller-%s-%s` 、 `cpu-model-history-length`、`24h` etc.?

Yes, we can do it later

pkg/autoscaling/estimator/percentile.go
xieydd: got it.

PR #285, opened 2022-04-26
add workload selector for evpa

### LINKED ISSUE #284: Use workload selector to select pods metrics when use metric server
## Describe the feature
Use workload selector to select pods metrics when use metric server

kitianFresh:

Fixes #284


pkg/prediction/dsp/prediction.go
```diff
@@ -30,9 +30,11 @@ const (
 
 type periodicSignalPrediction struct {
 	prediction.GenericPrediction
-	a           aggregateSignals
-	stopChMap   sync.Map
-	modelConfig config.AlgorithmModelConfig
+	a         aggregateSignals
+	stopChMap sync.Map
+	// record the query routine already started
+	queryRoutines sync.Map
```
yufeiyu, line 36: When does this map get set?

kitianFresh: Before we can not update the percentile algorithm config after start predicting, because general prediction will drop it  when call `WithQuery` when it is already registered, but we want to update the algorithm config of the metric dynamically.

so i moved the querySet to the algorithm internal, when user  call `WithQuery`, it can update the model config but do not started a  query go routine again

kitianFresh: > When does this map get set?

Before we can not update the percentile algorithm config after start predicting, because general prediction will drop it  when call `WithQuery` when it is already registered, but we want to update the algorithm config of the metric dynamically.

so i moved the querySet to the algorithm internal, when user  call `WithQuery`, it can update the model config but do not started a  query go routine again;

you can see the `line:166` is what i changed

pkg/autoscaling/estimator/percentile.go
```diff
@@ -66,28 +65,71 @@ func (e *PercentileResourceEstimator) GetResourceEstimation(evpa *autoscalingapi
...
+				Selector:      selector,
 			},
 		},
 	}
-
 	memConfig := getMemConfig(config)
-	tsList, err = utils.QueryPredictedValues(e.Predictor, caller, memConfig, memoryMetricNamer)
+
+	var errs []error
+	// first register cpu & memory, or the memory will be not registered before the cpu prediction succeed
```
qmhu, line 75: Would you please make this commend more clearer.

pkg/autoscaling/estimator/percentile.go
```diff
@@ -66,28 +65,71 @@ func (e *PercentileResourceEstimator) GetResourceEstimation(evpa *autoscalingapi
...
+	if len(recommendResource) == 0 {
+		return recommendResource, fmt.Errorf("all resource predicted failed, predictErrs: %v, noValueErrs: %v", predictErrs, noValueErrs)
+	}
 
+	// at least one succeed
 	return recommendResource, nil
 }
 
 func (e *PercentileResourceEstimator) DeleteEstimation(evpa *autoscalingapi.EffectiveVerticalPodAutoscaler) {
+	selector, err := e.TargetFetcher.Fetch(&corev1.ObjectReference{
```
qmhu, line 124: Is this selector only needed for metric server or common for both prom and metric server ?

pkg/autoscaling/estimator/percentile.go
kitianFresh: > Would you please make this commend more clearer.

ok. originally, we call `utils.QueryPredictedValues` it will do register & predict. then it failed for cpu, then it will be always return directly, but the memory is not registered forever before the cpu is succeed. 

so, for each predicting, we should first to register what you need to predict, then do predicting. because evpa maybe has multiple metrics for predicting, so we first register all of them, then get predicted value by reconciling

I will make it clear more

pkg/autoscaling/estimator/percentile.go
kitianFresh: > Is this selector only needed for metric server or common for both prom and metric server ?

maybe for prom is useful too, because we select containers and pods name by regex in prometheus now, but maybe use kube -state-metrics  is more safe for selecting pods by kube_pod_labels

metrics server is only can get pods metrics and node metrics for default metric server, so if we want to get workload or container metrics, we must has label selector to list pods, or we will list all pods metrics each time add sample.

you can see line `pkg/providers/metricserver/rest_metric_client.go:156`

PR #290, opened 2022-04-28
Fixed Crane dashboard discount modification invalid

lbbniu:
Fixed Crane dashboard discount modification invalid
real url：http://xxxx/grafana/d/cluster-costs/cluster-cost-and-utilization-metrics?chunkNotFound&orgId=1&var-Discount=85

Fixes #


PatrickLai7528 (approved):
ok

PR #305, opened 2022-05-07
fix analytics and recommendation status update

hoojos:
#### What type of PR is this?
/kind bugfix

`analytics` and `recommendation` Controller can not update resource's status correctly,  because we should use `Client.Status().Update()` to update `status`, `Client.Update()` can not update `status` field.

Fixes #


qmhu:
Analytics status isn't a subresource.

PR #309, opened 2022-05-10
add wildcard support for evpa container policy, and update the status…

### LINKED ISSUE #308: Support container wildcard for evpa containerPolicies
## Describe the feature
support container wildcard for evpa.

kitianFresh:
… even if it is off mode, because we need to record the metrics

#### What type of PR is this?
1. support wildcard for evpa container
2. refine the metrics for evpa to configure grafana
3. fix some bug 

Fixes #308

### DIFF
--- a/pkg/controller/evpa/container_policy.go
+++ b/pkg/controller/evpa/container_policy.go
@@ -28,16 +28,19 @@ func (c *EffectiveVPAController) ReconcileContainerPolicies(evpa *autoscalingapi
 	recommendation = evpa.Status.Recommendation
 
 	rankedEstimators := RankEstimators(resourceEstimators)
-	for _, containerPolicy := range evpa.Spec.ResourcePolicy.ContainerPolicies {
-		// container scaling is disabled
-		if (containerPolicy.ScaleUpPolicy.ScaleMode != nil && *containerPolicy.ScaleUpPolicy.ScaleMode == vpatypes.ContainerScalingModeOff) ||
-			(containerPolicy.ScaleDownPolicy.ScaleMode != nil && *containerPolicy.ScaleDownPolicy.ScaleMode == vpatypes.ContainerScalingModeOff) {
-			continue
+	needReconciledContainers := make(map[string]autoscalingapi.ContainerResourcePolicy)
+	for _, container := range podTemplate.Spec.Containers {
+		for _, containerPolicy := range evpa.Spec.ResourcePolicy.ContainerPolicies {
+			if containerPolicy.ContainerName == "*" || containerPolicy.ContainerName == container.Name {
+				out := containerPolicy.DeepCopy()
+				out.ContainerName = container.Name
+				needReconciledContainers[container.Name] = *out
+			}
 		}
-
+	}
+	for container, containerPolicy := range needReconciledContainers {
 		// get current resource by pod template
-		// todo: support "*"
-		resourceRequirement, found := utils.GetResourceByPodTemplate(podTemplate, containerPolicy.ContainerName)
+		resourceRequirement, found := utils.GetResourceByPodTemplate(podTemplate, container)
 		if !found {
 			klog.Warningf("ContainerName %s not found", containerPolicy.ContainerName)
 			continue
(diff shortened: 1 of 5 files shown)


pkg/controller/evpa/effective_vpa_controller.go
```diff
@@ -172,30 +195,32 @@ func recordMetric(evpa *autoscalingapi.EffectiveVerticalPodAutoscaler, status *a
...
 
 		recommendMem := container.Target[v1.ResourceMemory]
 		currentMem := resourceRequirement.Requests[v1.ResourceMemory]
+		labels["resource"] = v1.ResourceMemory.String()
 		if currentMem.Cmp(recommendMem) > 0 {
 			// scale down
 			currCopy := currentMem.DeepCopy()
 			currCopy.Sub(recommendMem)
-			metrics.EVPAMemoryScaleDownMB.With(labels).Set(float64(currCopy.Value() / 1024 / 1024))
+			metrics.EVPAMemoryScaleDownMB.With(labels).Set(float64(currCopy.Value()))
```
qmhu, line 218: Use MB unit here.

pkg/controller/evpa/container_policy.go
```diff
@@ -28,23 +28,28 @@ func (c *EffectiveVPAController) ReconcileContainerPolicies(evpa *autoscalingapi
 	recommendation = evpa.Status.Recommendation
 
 	rankedEstimators := RankEstimators(resourceEstimators)
-	for _, containerPolicy := range evpa.Spec.ResourcePolicy.ContainerPolicies {
-		// container scaling is disabled
-		if (containerPolicy.ScaleUpPolicy.ScaleMode != nil && *containerPolicy.ScaleUpPolicy.ScaleMode == vpatypes.ContainerScalingModeOff) ||
-			(containerPolicy.ScaleDownPolicy.ScaleMode != nil && *containerPolicy.ScaleDownPolicy.ScaleMode == vpatypes.ContainerScalingModeOff) {
-			continue
+	needReconciledContainers := make(map[string]autoscalingapi.ContainerResourcePolicy)
```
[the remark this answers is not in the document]
kitianFresh: > why remove the checking before?

now we do not apply really. and if we exit when the mode is off, the metrics will never record and the metrics always is null. this is not expected.

we need the metrics even if it is Off mode.

the real action to scale in & scale out is the place we judge the evpa scaleMode

pkg/controller/evpa/effective_vpa_controller.go
```diff
@@ -172,30 +195,32 @@ func recordMetric(evpa *autoscalingapi.EffectiveVerticalPodAutoscaler, status *a
 
 		recommendCpu := container.Target[v1.ResourceCPU]
 		currentCpu := resourceRequirement.Requests[v1.ResourceCPU]
+		labels["resource"] = v1.ResourceCPU.String()
 		if currentCpu.Cmp(recommendCpu) > 0 {
 			// scale down
 			currCopy := currentCpu.DeepCopy()
 			currCopy.Sub(recommendCpu)
-			metrics.EVPACpuScaleDownMilliCores.With(labels).Set(float64(currCopy.MilliValue()))
+			metrics.EVPACpuScaleDownMilliCores.With(labels).Set(float64(currCopy.MilliValue()) / 1000.)
```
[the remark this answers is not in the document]
kitianFresh: > Millicores is more accurate here.

I see that other metrics such as kube-state-metrics is not have unit in metricnaming, it is just use as a label to denote the unit.  such as `cores` and `bytes`. for example, we use it to join with `kube_pod_container_resource_requests`. 

it is better for us to do metircs and dashboard

pkg/controller/evpa/effective_vpa_controller.go
kitianFresh: > Use MB unit here.

same reason as above, i think we should not use the unit here, keep the original bytes. and use label to denote the unit

qmhu: ok， please also change variable name.

PR #312, opened 2022-05-11
Use caddy to replace nginx

### LINKED ISSUE #235: Start dashboard failed when grafana not exist
**Describe the bug**

Start dashboard failed when grafana not exist

**Reproduce steps**

skip grafana deployment, when we install craned, it will failed.

**Expected behavior**

craned should running.

**Screenshots**

 - K8S Version: [e.g. 1.19]
 - Crane Version: [e.g. 0.1.0] v0.3.0

lbbniu:
#### What type of PR is this?

Fix bug

Solve that the grafana service is not created, and the dashboard cannot be started.

Fixes #235

### DIFF
--- a/deploy/craned/deployment.yaml
+++ b/deploy/craned/deployment.yaml
@@ -61,8 +61,8 @@ spec:
           imagePullPolicy: IfNotPresent
           name: dashboard
           volumeMounts:
-            - mountPath: /etc/nginx/conf.d/
-              name: nginx-conf
+            - mountPath: /etc/caddy/
+              name: caddy-conf
       volumes:
         - name: cert
           secret:
@@ -71,9 +71,9 @@ spec:
         - name: config
           configMap:
             name: recommendation-config
-        - name: nginx-conf
+        - name: caddy-conf
           configMap:
-            name: nginx-conf
+            name: caddy-conf
 
 ---
 apiVersion: v1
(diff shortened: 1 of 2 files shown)


zsnmwy:
I think Nginx is more common than Caddy. 
Better for people to reference.

lbbniu:
The service used by upstream in nginx is not created, nginx cannot be started, is there a better way to start nginx?

zsnmwy:
> The service used by upstream in nginx is not created, nginx cannot be started, is there a better way to start nginx?

Is it possible to provide a sample for this case?

lbbniu:
> > The service used by upstream in nginx is not created, nginx cannot be started, is there a better way to start nginx?
> 
> Is it possible to provide a sample for this case?
https://github.com/gocrane/crane/issues/235
Do not deploy grafana, directly deploy crane

zsnmwy:
We can use var to avoid Nginx starting to fail when upstream fails.
But it needs to point DNS to the K8S cluster.

```nginx
server {
    # this is the internal DNS, the cache only for the 30s
    resolver 127.0.0.11 valid=30s;
    
    location ^~ /api/ {
        # other config entries omitted for breavity
    
        set $upstream api.awesome.com:9000;
 
        # nginx will now start if host is not reachable
        fastcgi_pass    $upstream; 
        fastcgi_index   index.php;
    }
}
```
Ref: https://sandro-keil.de/blog/let-nginx-start-if-upstream-host-is-unavailable-or-down/

lbbniu:
This method is not reliable, how to dynamically obtain the dns address?

lbbniu:
> We can use var to avoid Nginx starting to fail when upstream fails. But it needs to point DNS to the K8S cluster.
> 
> ```nginx
> server {
>     # this is the internal DNS, the cache only for the 30s
>     resolver 127.0.0.11 valid=30s;
>     
>     location ^~ /api/ {
>         # other config entries omitted for breavity
>     
>         set $upstream api.awesome.com:9000;
>  
>         # nginx will now start if host is not reachable
>         fastcgi_pass    $upstream; 
>         fastcgi_index   index.php;
>     }
> }
> ```
> 
> Ref: https://sandro-keil.de/blog/let-nginx-start-if-upstream-host-is-unavailable-or-down/

This method is not reliable, how to dynamically obtain the dns address?

lbbniu:
And svc should be written as a complete address. If the default domain name of the deployed machine is not cluster.local, how can it be changed to obtain the cluster domain name? , and the configuration is mounted through cm? How to dynamically generate cm content? @zsnmwy

zsnmwy:
> And svc should be written as a complete address. If the default domain name of the deployed machine is not cluster.local, how can it be changed to obtain the cluster domain name? , and the configuration is mounted through cm? How to dynamically generate cm content? @zsnmwy

I agree with you.
It's hard to make DNS correct in different clusters.
Nginx seems hard to deal with that when upstream fail.

PR #315, opened 2022-05-14
Add metric ext_cpu_total_distribute for local state

shijieqin:
add metric ext_cpu_total_distribute for local state

### DIFF
--- a/pkg/ensurance/collector/collector.go
+++ b/pkg/ensurance/collector/collector.go
@@ -14,6 +14,7 @@ import (
 	"github.com/gocrane/crane/pkg/common"
 	"github.com/gocrane/crane/pkg/ensurance/collector/cadvisor"
 	"github.com/gocrane/crane/pkg/ensurance/collector/nodelocal"
+	"github.com/gocrane/crane/pkg/ensurance/collector/noderesource"
 	"github.com/gocrane/crane/pkg/ensurance/collector/types"
 	"github.com/gocrane/crane/pkg/features"
 	"github.com/gocrane/crane/pkg/known"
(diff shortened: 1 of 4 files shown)


pkg/ensurance/collector/collector.go
```diff
@@ -179,6 +180,13 @@ func (s *StateCollector) UpdateCollectors() {
 		if _, exists := s.collectors.Load(types.CadvisorCollectorType); !exists {
 			s.collectors.Store(types.CadvisorCollectorType, cadvisor.NewCadvisorCollector(s.podLister, s.GetCadvisorManager()))
```
yan234280533, line 181: If the node resource feature disabled,  we can not collect the metric, can we add a judge for it

pkg/ensurance/collector/noderesource/noderesource.go
```diff
@@ -0,0 +1,63 @@
...
+		podLister:  podLister,
+	}
+}
+
+func (n *NodeResource) GetType() types.CollectType {
+	return types.NodeResourceCollectorType
+}
+
+func (n *NodeResource) Collect() (map[string][]common.TimeSeries, error) {
+	klog.V(4).Infof("NodeResourceCollector Collect")
```
yan234280533, line 38: Please modify the log level to 6, to avoid to print to much logs

pkg/ensurance/collector/noderesource/noderesource.go
```diff
@@ -0,0 +1,63 @@
...
+	node, err := n.nodeLister.Get(n.nodeName)
+	if err != nil {
+		return nil, err
+	}
+	pods, err := n.podLister.List(labels.Everything())
+	if err != nil {
+		return nil, err
+	}
+
+	allExtCpu := node.Status.Allocatable.Name(corev1.ResourceName(fmt.Sprintf(utils.ExtResourcePrefixFormat, corev1.ResourceCPU.String())), resource.DecimalSI).MilliValue()
```
yan234280533, line 48: if the Allocatable not have extend resource, when does the agent start, it will crash or not?

shijieqin: if the Allocatable not have extend resource, node.Status.Allocatable.Name returns an empty Quantity, it will not crash.

shijieqin:
> Can you give an example for how to use the "ext_cpu_total_distribute" metric? @shijieqin

I added a NodeQOSEnsurancePolicy called disablescheduling-when-ext-cpu-total-distribute
In examples.

PR #329, opened 2022-05-21
fix(docs): fix list style

zsnmwy:


zsnmwy:
/LGTM


before
image


after
image

PR #330, opened 2022-05-23
Add targetUtilization for percentile prediction

yufeiyu:
1. Add targetUtilization for percentile prediction
2. Refactor percentile prediction

#### What type of PR is this?
feat

Add targetUtilization for percentile prediction

Fixes #


pkg/recommend/advisor/resource_request.go
```diff
@@ -74,19 +79,23 @@ func makeMemConfig(props map[string]string) *config.Config {
 	if !exists {
 		marginFraction = "0.15"
 	}
-
+	targetUtilization, exists := props["resource.mem-target-utilization"]
+	if !exists {
+		targetUtilization = "1.0"
```
qmhu, line 84: `
          replicas.min-cpu-target-utilization: "30"
          replicas.max-cpu-target-utilization: "75"
          replicas.cpu-target-utilization: "50"
`

currently xxx-utilization is a 100 percent based float, better to align the same format for new configuration.

also please add it in deploy/craned/deployment.yaml and also the helm chart.

qmhu:
`
shouldAnalytics := c.ShouldAnalytics(analytics)
	if !shouldAnalytics {
		klog.V(4).Infof("Nothing happens for Analytics %s", req.NamespacedName)
		return ctrl.Result{}, nil
	}
`
Do you want to include the fix for this pr?

qmhu:
`
shouldAnalytics := c.ShouldAnalytics(analytics)
	if !shouldAnalytics {
		klog.V(4).Infof("Nothing happens for Analytics %s", req.NamespacedName)
		return ctrl.Result{}, nil
	}
`
Do you want to include the fix for this pr?

PR #331, opened 2022-05-23
Add docs and sample for evpa

qmhu:
#### What type of PR is this?

Feature

Add docs and examples for evpa

Fixes #


zouyee:
Does evpa only use default queries, how does it support rawquery, and does it split the calculation module to using the time prediction module?

PR #334, opened 2022-05-25
fix: crane dashboard crashed without prometheus service

PatrickLai7528:
#### What type of PR is this?
fix: crane dashboard crashed without prometheus service

Fixes #

### DIFF
--- a/deploy/craned/deployment.yaml
+++ b/deploy/craned/deployment.yaml
@@ -165,9 +165,7 @@ data:
         server craned.crane-system:8082;
     }
 
-    upstream grafana {
-        server grafana.crane-system:8082;
-    }
+    resolver kube-dns.kube-system.svc.cluster.local valid=5s;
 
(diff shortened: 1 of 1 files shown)


lbbniu:
What if the domain name of the k8s cluster is not cluster.local ?

zsnmwy:
Maybe we can use Lua to get the nameserver address in `/etc/resolv.conf` when Nginx is in bootstrap.

```bash
/ # cat /etc/resolv.conf
search default.svc.cluster.local svc.cluster.local cluster.local
nameserver 10.43.0.10
options ndots:5
```

We can also get cluster domain in `/etc/resolv.conf`.
```bash
/ # awk '/search/{print $2}' /etc/resolv.conf
default.svc.cluster.local
```

If it fail, then the upstream fall back to `kube-dns.kube-system.svc.cluster.local`.

PR #348, opened 2022-06-06
Watch configSet file and update config for recommendation on the fly.

yufeiyu:
#### What type of PR is this?
feature

Watch configSet file and update config for recommendation on the fly.

Fixes #


qmhu:
Prometheus use a sidecar to reload configmap change, can we use it?

PR #429, opened 2022-07-25
fix(web): remove unnecessary grafana dashboard query str

PatrickLai7528:
#### What type of PR is this?
Remove unnecessary grafana dashboard query str

In some cases, adding ?query=% to /grafana/api/search will cause an error.
image

And I have tried that everything is working fine after removing ?query=%

Fixes #


zsnmwy (approved):
/LGTM

PR #450, opened 2022-08-03
bugfix:config change for example:workload-replicas

qmhu:
Signed-off-by: qmhu <brillantroad@gmail.com>

#### What type of PR is this?

bugfix

config change for example:workload-replicas

Fixes #


zsnmwy (approved):
/LGTM

PR #479, opened 2022-08-15
always collect node local metrics when node resource controller is enabled.

mfanjie:
#### What type of PR is this?
Feature

Currently, node resource controller is enabled by default, but node local agent metrics collector is not, which cause a repeated error in crane agent.

Fixes #


chenkaiyue (approved):
lgtm

PR #481, opened 2022-08-16
adapt to nodeqos and podqos api

mfanjie:

Fixes #


pkg/ensurance/executor/podinfo/pod_info.go
```diff
@@ -0,0 +1,138 @@
...
+	ElasticCPU                                                                      int64
+	PodCPUUsage, PodCPUShare, PodCPUQuota, PodCPUPeriod                             float64
+	ContainerCPUUsages, ContainerCPUShares, ContainerCPUQuotas, ContainerCPUPeriods []ContainerState
+
+	ActionType  ActionType
+	CPUThrottle CPURatio
+	Executed    bool
+}
+
+func ContainsPendingPod(pods []PodContext) bool {
```
chenkaiyue, line 97: Pending? What about ContainsNoExecutedPod?

pkg/resource/node_resource_manager.go
```diff
@@ -130,7 +130,6 @@ func (o *NodeResourceManager) Run(stop <-chan struct{}) {
 			case state := <-o.stateChann:
 				o.state = state
 				o.lastStateTime = time.Now()
-			case <-tspUpdateTicker.C:
```
chenkaiyue, line 133: Why delete this case ?

chenkaiyue:
https://github.com/mfanjie/crane/pull/1

pkg/ensurance/executor/podinfo/pod_info.go
mfanjie: pending means pending on actions to be executed, but it is confused cause pending generally means pending on scheduling in k8s, let me rename it.

pkg/resource/node_resource_manager.go
mfanjie: cause the origin logic is
1. collect the state from node by 10s interval
2.  compute node resource by 20s interval

The map is not synced, so the agent crashes frequently with `concurrent access to the map` error. 
By thinking deeper, I think node resource controller do not need to trigger the reconcile by interval, as long as there is new message from the channel, it should do reconcile. this simplifies  the flow.

chenkaiyue (approved):
lgtm

PR #514, opened 2022-08-22
fix craned deployment yaml

szpnygo:
Fixes  error: error parsing deploy/craned/deployment.yaml: error converting YAML to JSON: yaml: line 62: mapping values are not allowed in this context


qmhu (approved):
/LGTM

PR #526, opened 2022-08-24
optimize

szpnygo:

Fixes #

### DIFF
--- a/pkg/utils/slice.go
+++ b/pkg/utils/slice.go
@@ -10,14 +10,12 @@ func ContainsString(slice []string, str string) bool {
 }
 
 func RemoveString(slice []string, str string) []string {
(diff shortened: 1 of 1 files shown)


pkg/utils/slice.go
```diff
@@ -10,14 +10,12 @@ func ContainsString(slice []string, str string) bool {
...
 
 func RemoveString(slice []string, str string) []string {
-	if len(slice) == 0 {
-		return slice
-	}
-	var newSlice []string
-	for _, item := range slice {
-		if item != str {
-			newSlice = append(newSlice, item)
+	for index, item := range slice {
```
qmhu, line 13: Would you please add a unit test for it

PR #531, opened 2022-08-26
bugfix: fix rr selector error

qmhu:
#### What type of PR is this?

bugfix

fix the recommendation rule select error

Fixes #


xieydd (changes requested):
/lgtm

we can add unit test for this.

PR #547, opened 2022-09-08
make function GetPodTemplate and IsHPAControlledByEHPA more precise

whitebear009:
#### What type of PR is this?
optimize

In the function GetPodTemplate, validates the apiVersion and turns "group/version" string into a GroupVersion struct so they could be exactly compared instead of strings.HasPrefix. The latter may be misjudged, like apps.kruise.io which also have StatefulSet.
In hpa_event_handler.go, use ownerRefenerce to determine if hpa is controlled by ehpa instead of strings.HasPrefix.

### DIFF
--- a/pkg/controller/ehpa/hpa_event_handler.go
+++ b/pkg/controller/ehpa/hpa_event_handler.go
@@ -1,8 +1,6 @@
 package ehpa
 
 import (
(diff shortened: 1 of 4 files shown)


qmhu (approved):
/LGTM

PR #552, opened 2022-09-15
Watermark of cpu usage percent

chenkaiyue:
Watermark of cpu usage percent


chenkaiyue:
企业微信截图_e371a8ac-125d-4848-a838-2e68d9a2351a

pkg/utils/pod.go
```diff
@@ -174,6 +199,18 @@ func GetContainerNameFromPod(pod *v1.Pod, containerId string) string {
...
 	}
 
+	// for containerd
+	for _, v := range pod.Status.ContainerStatuses {
+		strList := strings.Split(v.ContainerID, "//")
+		if len(strList) > 0 {
+			klog.V(6).Infof("cri-containerd is %s ", "cri-containerd-"+strList[len(strList)-1]+".scope")
+			klog.V(6).Infof("containerid is %s", containerId)
+			if "cri-containerd-"+strList[len(strList)-1]+".scope" == containerId {
+				klog.V(6).Infof("111111111")
```
mfanjie, line 209: you printed 111111 here

pkg/utils/pod.go
```diff
@@ -174,6 +199,18 @@ func GetContainerNameFromPod(pod *v1.Pod, containerId string) string {
 		}
 	}
 
+	// for containerd
+	for _, v := range pod.Status.ContainerStatuses {
+		strList := strings.Split(v.ContainerID, "//")
+		if len(strList) > 0 {
+			klog.V(6).Infof("cri-containerd is %s ", "cri-containerd-"+strList[len(strList)-1]+".scope")
+			klog.V(6).Infof("containerid is %s", containerId)
+			if "cri-containerd-"+strList[len(strList)-1]+".scope" == containerId {
```
mfanjie, line 208: "cri-containerd-"+strList[len(strList)-1]+".scope"
change to fmt.Sprintf?

mfanjie: and did you use the same method from kubelet to find the containerId?

pkg/ensurance/executor/cpu_usage_percent.go
```diff
@@ -0,0 +1,70 @@
...
+	ThrottleFunc:       nil,
+	RestoreFunc:        nil,
+
+	Evictable:       true,
+	EvictQuantified: true,
+	EvictFunc:       cpuUsagePercentEvictPod,
+}
+
+func cpuUsagePercentEvictPod(wg *sync.WaitGroup, ctx *ExecuteContext, index int, totalReleasedResource *ReleaseResource, EvictPods EvictPods) (errPodKeys []string, released ReleaseResource) {
+	wg.Add(1)
```
mfanjie, line 35: seems you did not call wg.wait in main thread

pkg/utils/pod.go
chenkaiyue: OK，I didn't found codes related in kubelet, there is only some code like findContainerIDByName. And i search in github, https://github.com/search?p=1&q=cri-containerd-+.scope&type=Code , many projects do like us.

pkg/ensurance/executor/cpu_usage_percent.go
chenkaiyue: We call it in crane/pkg/ensurance/executor/evict.go:113

mfanjie (approved):
LGTM

PR #568, opened 2022-09-26
fix prediction without resource metrics

saikey0379:
#### What type of PR is this?
Bug of IsEHPAHasPredictionMetric

fix prediction without resource metrics

Fixes #


qmhu (approved):
/LGTM

PR #584, opened 2022-10-16
feat: use CoreDNS instead of kube-dns

szpnygo:
CoreDNS became the default DNS service for Kubernetes 1.13+ onwards.

btw, we can set the dns service as the argument in helm

### DIFF
--- a/deploy/craned/deployment.yaml
+++ b/deploy/craned/deployment.yaml
@@ -207,7 +207,7 @@ data:
         server craned.crane-system:8082;
     }
 
-    resolver kube-dns.kube-system.svc.cluster.local valid=5s;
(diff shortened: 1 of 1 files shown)


mfanjie:
kube-dns is still the default service name not coredns.

szpnygo:
Sorry, I look that the svc of my other clusters is kube-dns. The cluster svc I installed with kubekey yesterday is coredns. I'm looking for the reason in kubekey

szpnygo:
https://kubernetes.io/docs/tasks/administer-cluster/dns-debugging-resolution/#is-dns-service-up

I think this is a bug for kubekey

mfanjie:
@szpnygo look forward for more contributions~

PR #620, opened 2022-11-20
feature:support configurable cpu/memory ratio for resource recommendation

Mokaz111:
#### What type of PR is this?
feature

The result for resource recommendation don't consider the cpu/memory ratio now, It would be better to support configurable cpu/memory ratio for resource recommendation.

Fixes #
issue#612

### DIFF
--- a/pkg/recommendation/recommender/resource/recommend.go
+++ b/pkg/recommendation/recommender/resource/recommend.go
@@ -131,6 +131,13 @@ func (rr *ResourceRecommender) Recommend(ctx *framework.RecommendationContext) e
 		memQuantity := resource.NewQuantity(v, resource.BinarySI)
 		cr.Target[corev1.ResourceMemory] = memQuantity.String()
 
+		//use memory-to-vCPU-ratio if exist
+		if rr.MemTovCPURatioBool {
+			cpu, mem := rr.getVMSpec(cpuQuantity, memQuantity)
+			*cpuQuantity = cpu
+			*memQuantity = mem
+		}
+
 		newContainerSpec := corev1.Container{
 			Name: c.Name,
 			Resources: corev1.ResourceRequirements{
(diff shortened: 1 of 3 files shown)


pkg/recommendation/recommender/resource/recommend.go
```diff
@@ -229,3 +236,47 @@ func (rr *ResourceRecommender) Recommend(ctx *framework.RecommendationContext) e
 func (rr *ResourceRecommender) Policy(ctx *framework.RecommendationContext) error {
 	return nil
 }
+
+var cvmSpecList []float64
+var cvmSpecMap map[float64][]float64
+
+func (rr *ResourceRecommender) getVMSpec(cpu, mem *resource.Quantity) (resource.Quantity, resource.Quantity) {
```
qmhu, line 243: Better to add a unit test for function getVMSpec

pkg/recommendation/recommender/resource/recommend.go
```diff
@@ -131,6 +131,13 @@ func (rr *ResourceRecommender) Recommend(ctx *framework.RecommendationContext) e
 		memQuantity := resource.NewQuantity(v, resource.BinarySI)
 		cr.Target[corev1.ResourceMemory] = memQuantity.String()
 
+		//use memory-to-vCPU-ratio if exist
+		if rr.MemTovCPURatioBool {
+			cpu, mem := rr.getVMSpec(cpuQuantity, memQuantity)
```
qmhu, line 136: if configuration exist and pass the format checking, then call rr.getVMSpec.
Also we should add a log here to know it happens.

pkg/recommendation/recommender/resource/registry.go
```diff
@@ -73,6 +75,15 @@ func NewResourceRecommender(recommender apis.Recommender) *ResourceRecommender {
 	if !exists {
 		memHistoryLength = "168h"
 	}
+	memTovCPURatio, exists := recommender.Config["memory-to-vCPU-ratio"]
```
qmhu, line 78: As we discuss, the configuration key should be "resource-specification"

PR #642, opened 2022-12-06
fix: remove duplicated reset triggered map

### LINKED ISSUE #637: QOS: ensurance analyzer: Reset the element of triggered map duplicately when compute the action context
**Describe the bug**
the demo is here  : pkg/ensurance/analyzer/analyzer.go:286
image

**Reproduce steps**
1. analyze the node QOS (u might call it ensurance):
2. if aboveThreshold is true —— it is undoubtfully to reset restored map where key = nodeName + ruleName. it means the Restore Process(please allow me to call it) is broken off once the aboveThreshold is true.
3. if aboveThreshold is false —— it might mean the node is in the <Restore Processing> . In the future,  the node would be judged as <Restored = true> when the restore times accumulate to object.RestoreThreshold. At that time, the element of triggered map would be reset to zero , meaning the node could be forgiven and be A restored node again **ONLY IF THE CONDITION (s.triggered[key] >= uint64(object.AvoidanceThreshold) &&  s.restored[key] >= uint64(object.RestoreThreshold)) IS TRUE.**
In summary , i think the demo in (pkg/ensurance/analyzer/analyzer.go:286) should be deleted, meanwhile the demo in (pkg/ensurance/analyzer/analyzer.go:298) remains.

**Expected behavior**
the duplicated reset step would lead to the condition (pkg/ensurance/analyzer/analyzer.go:295) be meaningless forever.
**Screenshots**
image

 - K8S Version: [e.g. 1.19]
 - Crane Version: [e.g. 0.1.0]

Sincerely.

shanyuyulaifengmanlou:
#### What type of PR is this?
For QoS-Ensurance bug fixing.

Removing the duplicated action of reseting the triggered map would fix the bug of RESTORE IF-CONDITION(pkg/ensurance/analyzer/analyzer.go:295) never work.

Fixes #637 

#### Special notes for your reviewer:
Sincerely

### DIFF
--- a/pkg/ensurance/analyzer/analyzer.go
+++ b/pkg/ensurance/analyzer/analyzer.go
@@ -283,7 +283,6 @@ func (s *AnomalyAnalyzer) computeActionContext(aboveThreshold bool, key string, 
 			ac.Triggered = true
 		}
 	} else {
-		s.triggered[key] = 0
 		restored := utils.GetUint64FromMaps(key, s.restored)
 		restored++
 		// log how many times the actual usage below threshold


chenkaiyue (approved):
lgtm

PR #648, opened 2022-12-13
Override recommder parameter with the config in RecommendationRule

whitebear009:
#### What type of PR is this?
optimization

Override the RecommendationConfiguration in configmap and the default value with the config in RecommendationRule
	
Fixes #


pkg/recommendation/recommender/resource/registry.go
```diff
@@ -36,118 +54,137 @@ func (rr *ResourceRecommender) Name() string {
...
 		OOMBumpRatioFloat,
-		specificationBool,
-		specifications,
-	}, nil
+		defaultSpecificationBool,
+		defaultSpecifications,
+	}
+
+	// Override the default value with RecommendationConfiguration in configmap
+	if err = SetRecommendationConfig(rr, recommender.Config); err != nil {
```
qmhu, line 182: 1. merge recommender.Config and recommendationRule.Spec.Config first
2. do other initialization steps

whitebear009: 1. if the value in recommender.Config and recommendationRule.Spec.Config conflict, I will use the value of recommendationRule. Is that right?

qmhu: > 1. if the value in recommender.Config and recommendationRule.Spec.Config conflict, I will use the value of recommendationRule. Is that right?

yes, recommendationRule.Spec.Config have the highest priority.

qmhu: also, api change PR should be merged first and use the newest api version in crane

whitebear009: ok

qmhu:
https://github.com/gocrane/api/pull/74 has merged, and you can run `make all` locally to avoid CI error

whitebear009:
@qmhu  it's done. please check.

pkg/recommendation/recommender/idlenode/registry.go
```diff
@@ -21,10 +21,19 @@ func (inr *IdleNodeRecommender) Name() string {
 }
 
 // NewIdleNodeRecommender create a new idle node recommender.
-func NewIdleNodeRecommender(recommender apis.Recommender) (*IdleNodeRecommender, error) {
+func NewIdleNodeRecommender(recommender apis.Recommender, recommendationRuleConfig map[string]string) (*IdleNodeRecommender, error) {
 	if recommender.Config == nil {
 		recommender.Config = map[string]string{}
 	}
```
qmhu, line 27: Could you extract these line into a common function?

whitebear009:
/hold 
I found something wrong.

qmhu (approved):

PR #656, opened 2022-12-16
change event type when meet error

whitebear009:
#### What type of PR is this?
optimization

change event type when meet error

Fixes #


qmhu (approved):
good catch!

PR #671, opened 2023-01-04
fix the panic of concurrent map write

yingnanzhang666:
#### What type of PR is this?
Bug Fix. 

fix the panic of concurrent map write at
https://github.com/gocrane/crane/blob/main/pkg/recommend/config.go#L67

Fixes #


qmhu (approved):

qmhu:
Great job, welcome @yingnanzhang666

PR #677, opened 2023-01-13
fix default histogram MaxValue of makeMemConfig

RTsien:
#### What type of PR is this?
fix

adjust the percentile recommended memory maximum value from 100Gi to 300Gi to to support large memory workload

Fixes #


qmhu:
Could you test the memory increase for this change? I will bring three times percentile structs in memory.

qmhu:
closed caused by https://github.com/gocrane/crane/pull/699

PR #681, opened 2023-01-17
support reserved cpus

xrmzju:
#### What type of PR is this?
enhancements

Fixes # [680](https://github.com/gocrane/crane/issues/680)

Fixes # [680](https://github.com/gocrane/crane/issues/680)


pkg/ensurance/cm/cpumanager/cpu_manager.go
```diff
@@ -631,11 +639,32 @@ func fromZoneListToTopologyResult(zones topologyapi.ZoneList) (TopologyResult, e
...
+		return reservedSystemCPUs
+	}
+	return cpuset.NewCPUSet()
+}
+
+func shouldExcludeReservedCPUs(pod *corev1.Pod) bool {
+	if pod == nil {
+		return false
+	}
+	return pod.Annotations[topologyapi.AnnotationPodExcludeReservedCPUs] == "true"
```
xieydd, line 669: need check the key exist in map.

xieydd (changes requested):

pkg/ensurance/cm/cpumanager/cpu_manager.go
```diff
@@ -631,11 +639,32 @@ func fromZoneListToTopologyResult(zones topologyapi.ZoneList) (TopologyResult, e
...
+			// invalid cpu list is provided, set reservedSystemCPUs to empty, so it won't overwrite kubeReserved/systemReserved
+			klog.Errorf("Invalid ReservedSystemCPUs \"%s\"", reservedSystemCPUsStr)
+		}
+		return reservedSystemCPUs
+	}
+	return cpuset.NewCPUSet()
+}
+
+func shouldExcludeReservedCPUs(pod *corev1.Pod) bool {
+	if pod == nil {
```
Garrybest, line 666: nit: actually we don't need to check because all pods listed from kubelet must not be nil

pkg/ensurance/collector/noderesourcetopology/noderesourcetopology.go
```diff
@@ -120,6 +120,12 @@ func BuildNodeResourceTopology(sysPath string, kubeletConfig *kubeletconfiginter
 	nrtBuilder.WithReservedCPUs(getNumReservedCPUs(reserved))
```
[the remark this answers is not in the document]
Garrybest: Also we need to update allocatable resources for all numa nodes, which have a impact on scheduling. If `--reserved-cpus` is specified, I guess all allocatable cpu resource here is incorrect.

pkg/ensurance/cm/cpumanager/cpu_manager.go
Garrybest: I think there is no need to check. If a map is nil or the key does not exist, we will get an empty string here which could never be equal to "true".

xieydd: got it.

pkg/ensurance/cm/cpumanager/cpu_manager.go
xrmzju: just a protection for panic...

xrmzju:
@xieydd @Garrybest PTAL.

pkg/ensurance/collector/noderesourcetopology/noderesourcetopology.go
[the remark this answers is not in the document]
xrmzju: fixed

Garrybest:
Good job👍.

/lgtm

mfanjie (approved):
LGTM

Garrybest (approved):

xieydd (approved):
LGTM

PR #692, opened 2023-02-13
delete recommendation belong to recommendationrule if target oject not found

whitebear009:
#### What type of PR is this?
optimize

if we deploy recommendationrule to cover all deployment/statefulset like example, craned will create correspond recommendation for all deployment/statefulset. 
If we delete some workloads (required for testing), the recommendation will not be deleted. After a period of time, there will be many unavailable recommedations in the cluster.

so this PR optimizes the issue. if recommendation is created by recommendationrule and its target object has been deleted, delete correspond recommendation.

Fixes #


qmhu (commented):
When recommendationrule controller rerun the progress, the orphan recommendation also will be deleted?

qmhu (approved):

qmhu:
Solve the error that throws in RecommendationController, good improvement.

PR #693, opened 2023-02-14
observe metrics in recommendation controller

### LINKED ISSUE #672: recommendation metrics lost after restart craned
**Describe the bug**

```
crane_analysis_resource_recommendation
crane_analysis_resource_recommendation
...
```

**Reproduce steps**
1. start craend
2. recommend success once
3. confirm metrics exists
3. restart craned
4. metrics lost

**Expected behavior**

**Screenshots**

 - K8S Version: 1.22
 - Crane Version: 0.9.0

whitebear009:
#### What type of PR is this?
optimize

Currutly recommendation metrics will lost after restart craned because the metrics is created in the progress of calculation of recommend.
If we observe recommendation status in recommedation controller, that can avoid this issue. 

Fixes #672

### DIFF
--- a/cmd/craned/app/manager.go
+++ b/cmd/craned/app/manager.go
@@ -368,11 +368,12 @@ func initControllers(oomRecorder oom.Recorder, mgr ctrl.Manager, opts *options.O
 		}
 
 		if err := (&recommendationctrl.RecommendationController{
-			Client:      mgr.GetClient(),
-			Scheme:      mgr.GetScheme(),
-			RestMapper:  mgr.GetRESTMapper(),
-			ScaleClient: scaleClient,
-			Recorder:    mgr.GetEventRecorderFor("recommendation-controller"),
+			Client:         mgr.GetClient(),
+			Scheme:         mgr.GetScheme(),
+			RestMapper:     mgr.GetRESTMapper(),
+			RecommenderMgr: recommenderMgr,
+			ScaleClient:    scaleClient,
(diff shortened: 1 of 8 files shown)


pkg/metrics/analysis.go
```diff
@@ -13,7 +13,7 @@ var (
 			Name:      "resource_recommendation",
 			Help:      "The containers' CPU/Memory recommended value",
 		},
-		[]string{"apiversion", "owner_kind", "namespace", "owner_name", "container", "resource", "owner_replicas"},
```
zxh326, line 16: `owner_replicas`
Why should we discard this label？

whitebear009: Because I thought owner_replicas is not a value that resource recommend needs to care about, there is also no related value in recommendation CRD. 

I refactor thses code with a common method (the input parameter of the observe function was changed from RecommendationContext to Recommendation). If this field needs to be displayed, it needs to be processed separately. 

Please let me know if there is a need to keep this field or any other information I am missing.

qmhu: replicas * resource = total workload resource now
so i think it is useful here.

whitebear009: OK, I will adjust it later

whitebear009:
@qmhu  Already updated, PTAL

qmhu:
> @qmhu Already updated, PTAL

Please fix the lint

whitebear009:
Sorry, miss that. It is fixed now.

qmhu (approved):

PR #694, opened 2023-02-14
add maxConcurrentReconciles flag for tsp controller

whitebear009:
#### What type of PR is this?
optimize

Currently tsp controller reconcile a goroutine to query data from prometheus. Sometimes query take long time and need to increase the value of goroutines to run the reconcile.

Fixes #

### DIFF
--- a/cmd/craned/app/manager.go
+++ b/cmd/craned/app/manager.go
@@ -347,7 +347,7 @@ func initControllers(oomRecorder oom.Recorder, mgr ctrl.Manager, opts *options.O
 			predictorMgr,
 			targetSelectorFetcher,
 		)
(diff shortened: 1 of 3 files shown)


qmhu (commented):
Is is only for tsp controller?

whitebear009:
At present, I think the performance problem of tsp is relatively big(calculating custom metrics maybe very slow), so I added it to tsp first (I see kube
-controller-manager is also a separate flag for each controller, https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/#options).

There may be some concurrency problems when increasing the number of concurrency. I haven't tested other controllers yet. After the test, I can also add flags to other controllers. How do you think?

qmhu (approved):
LGTM

PR #696, opened 2023-02-17
fix analytics webhook function

champly:
#### What type of PR is this?

`analysisapi.Analytics` webhook should use `pkg/webhook/analytics`

Fixes #


qmhu (commented):
`RecommendationRule` is the replacement for `Analaytics` , so it's ok here.

champly:
> 

ok, then i close this pr.

PR #714, opened 2023-03-09
feat: Support trigger a recommendation manually

qmhu:
#### What type of PR is this?

Feature

Manually triggering a recommendation can be done through the following steps:

1. Obtain the runNumber from the recommendationRule.
image

2. Modify the run number annotation for the recommendation that should be triggered. The run number should be lower than the runNumber in the recommendationRule. For example, if recommendationRule.status.runNumber=6, then change the annotation value to 5.
image

3. The recommendationRule controller checks for updates every 10 seconds. Once it detects the modified run number, it triggers the recommendation flow and updates the recommendationRule.status and recommendation. The recommendation's runNumber is modified to match that of the recommendationRule. You can use this to know when the manual trigger is finished.

Fixes #

### DIFF
--- a/pkg/controller/recommendation/recommendation_rule_controller.go
+++ b/pkg/controller/recommendation/recommendation_rule_controller.go
@@ -41,18 +41,16 @@ import (
 
 type RecommendationRuleController struct {
 	client.Client
-	Scheme                   *runtime.Scheme
-	Recorder                 record.EventRecorder
-	RestMapper               meta.RESTMapper
-	ScaleClient              scale.ScalesGetter
-	RecommenderMgr           recommender.RecommenderManager
-	PredictorMgr             predictormgr.Manager
-	kubeClient               kubernetes.Interface
-	dynamicClient            dynamic.Interface
-	discoveryClient          discovery.DiscoveryInterface
-	Provider                 providers.History
-	RecommendationRuleStatus map[string]bool
-	lock                     sync.RWMutex
+	Scheme          *runtime.Scheme
+	Recorder        record.EventRecorder
+	RestMapper      meta.RESTMapper
+	ScaleClient     scale.ScalesGetter
+	RecommenderMgr  recommender.RecommenderManager
+	PredictorMgr    predictormgr.Manager
+	kubeClient      kubernetes.Interface
+	dynamicClient   dynamic.Interface
+	discoveryClient discovery.DiscoveryInterface
+	Provider        providers.History
 }
 
 func (c *RecommendationRuleController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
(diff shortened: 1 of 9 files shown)


pkg/controller/recommendation/recommendation_rule_controller.go
```diff
@@ -278,11 +286,108 @@ func (c *RecommendationRuleController) SetupWithManager(mgr ctrl.Manager) error
...
+	}()
+}
+
+func (c *RecommendationRuleController) setStatus(recommendationRule *analysisv1alph1.RecommendationRule, isRunning bool) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	c.RecommendationRuleStatus[string(recommendationRule.UID)] = isRunning
+}
+
+func (c *RecommendationRuleController) getStatus(recommendationRule *analysisv1alph1.RecommendationRule) (bool, bool) {
```
whitebear009, line 383: the same as above

whitebear009:
It seems that only a single recommendation can be manually executed. Is it necessary to support manual execution of the recommendationrule?

pkg/controller/recommendation/recommendation_rule_controller.go
```diff
@@ -278,11 +286,108 @@ func (c *RecommendationRuleController) SetupWithManager(mgr ctrl.Manager) error
...
+					}
+
+				}
+
+			}
+		}
+	}()
+}
+
+func (c *RecommendationRuleController) setStatus(recommendationRule *analysisv1alph1.RecommendationRule, isRunning bool) {
```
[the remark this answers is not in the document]
qmhu: remove it

pkg/controller/recommendation/recommendation_rule_controller.go
qmhu: also remove it

qmhu:
> 

Last implement has a limitation: if the recommendation rule controller is reconciling, the manual trigger will wait, this is unexpectable. So we use a trigger controller to watch the decrease event.

How to trigger a recommendation is trick work. We can't modify the spec, but recommendationRule controller only watch spec changes...

qmhu:
@whitebear009 any comments?

pkg/controller/recommendation/recommendation_trigger_controller.go
```diff
@@ -0,0 +1,160 @@
...
+func (h *recommendationEventHandler) Delete(evt event.DeleteEvent, q workqueue.RateLimitingInterface) {
+}
+
+func (h *recommendationEventHandler) Update(evt event.UpdateEvent, q workqueue.RateLimitingInterface) {
+	newRecommendation := evt.ObjectNew.(*analysisv1alpha1.Recommendation)
+	oldRecommendation := evt.ObjectOld.(*analysisv1alpha1.Recommendation)
+	klog.V(6).Infof("recommendation %s OnUpdate", klog.KObj(newRecommendation))
+
+	oldRunNumber, _ := utils.GetRunNumber(oldRecommendation)
+	newRunNumber, _ := utils.GetRunNumber(newRecommendation)
```
whitebear009, line 152: the error is not handled

qmhu: if get run number failed , we treat it as a empty number which by default is 0.

PR #724, opened 2023-03-24
add metric kind for prometheus builder

whitebear009:
#### What type of PR is this?
optimize

add metric kind for prometheus builder, otherwise in the GetPodNameReg method, the switch will not be entered.

```
func GetPodNameReg(resourceName string, resourceType string) string {
	switch resourceType {
	case "ReplicaSet":
		return fmt.Sprintf("^%s-%s", resourceName, PostRegMatchesPodReplicaset)
	case "Deployment":
		return fmt.Sprintf("^%s-%s", resourceName, PostRegMatchesPodDeployment)
	case "StatefulSet":
		return fmt.Sprintf("^%s-%s", resourceName, PostRegMatchesPodStatefulset)
	}
	return fmt.Sprintf("^%s-%s", resourceName, `.*`)

}
```


whitebear009:
@qmhu  PTAL，thanks

qmhu (commented):
Please take a look @saikey0379

saikey0379 (approved):

qmhu (approved):

PR #725, opened 2023-03-27
fix gitVersion in makefile

2456868764:
#### What type of PR is this?
make gitVersion problem in make file

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ ifeq ($(GIT_DIFF), 1)
 endif
 BUILDDATE = $(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
 
-LDFLAGS = "-X github.com/gocrane/crane/pkg/version.gitVersion=$(GIT_VERSION) \
+LDFLAGS = "-X github.com/gocrane/crane/pkg/version.gitTag=$(GIT_VERSION) \
                       -X github.com/gocrane/crane/pkg/version.gitCommit=$(GIT_COMMIT_HASH) \
                       -X github.com/gocrane/crane/pkg/version.gitTreeState=$(GIT_TREESTATE) \
                       -X github.com/gocrane/crane/pkg/version.buildDate=$(BUILDDATE)"


qmhu (commented):
Would you please provide more information about current problem and how it fix?

2456868764:
> Would you please provide more information about current problem and how it fix?

1. this is version.go code : 

`package version

import (
	"encoding/json"
	"fmt"
	"runtime"
)

var (
	gitVersion   = "crane-%s"
	gitCommit    = "$Format:%H$" // sha1 from git, output of $(git rev-parse HEAD)
	gitTreeState = ""            // state of git tree, either "clean" or "dirty"
	gitTag       = ""
	buildDate    = "1970-01-01T00:00:00Z" // build date in ISO8601 format, output of $(date -u +'%Y-%m-%dT%H:%M:%SZ')
)

type Version struct {
	GitVersion   string `json:"gitVersion"`
	GitCommit    string `json:"gitCommit"`
	BuildDate    string `json:"buildDate"`
	GoVersion    string `json:"goVersion"`
	Compiler     string `json:"compiler"`
	Platform     string `json:"platform"`
	GitTreeState string `json:"gitTreeState"`
}

func GetVersionInfo() string {
	ver := Version{
		GitVersion:   fmt.Sprintf(gitVersion, gitTag),
		GitCommit:    gitCommit,
		BuildDate:    buildDate,
		GoVersion:    runtime.Version(),
		Compiler:     runtime.Compiler,
		Platform:     fmt.Sprintf("%s/%s", runtime.GOOS, runtime.GOARCH),
		GitTreeState: gitTreeState,
	}
	res, _ := json.Marshal(ver)
	return string(res)
}`

2. according to code ： ”GitVersion:   fmt.Sprintf(gitVersion, gitTag),“ 
so gitVersion will be replaced by the current gitTag when it is built, and not started with "crane-".

To fix this bug, gitTag will be replaced when it's built , and gitVersion will be started with "crane-" .

qmhu (approved):

PR #727, opened 2023-03-30
add option for cache unstructured

whitebear009:
#### What type of PR is this?
optimization

Currently craned use unstructured object in many scenarios. 
The controller-runtime client does not use cache when reading unstructured, which may cause slow performance, but enabling cache will increase memory usage.
So an option is added to allow users to choose whether to enable.


whitebear009:
@qmhu PTAL, thanks

qmhu (approved):
LGTM

Issue #754, opened 2023-05-04
v0.10.0 crane dashboard "Add Cluster view" the required field validation prompt is incorrect


JaneConan:
**Describe the bug**
"Add Cluster view" the required field validation prompt is incorrect.

if we change Discount value < 0

**Reproduce steps**

1. open the "Add Cluster view"
2. click "Confirm" button
3. show the prompted.

**Expected behavior**

**Screenshots**
![image](https://user-images.githubusercontent.com/26090153/236146024-4bf44d17-4737-4317-8ab9-67b4da8a2fcf.png)
![image](https://user-images.githubusercontent.com/26090153/236146422-89827698-260a-485a-8d8a-6f49ca5402b8.png)

**Environment (please complete the following information):**
 - K8S Version: [e.g. 1.19]
 - Crane Version: [e.g. 0.1.0]

PR #760, opened 2023-05-04
fix:修复bug 折扣为负数的问题 并且设置了折扣的取值范围

RuoChengF:
针对 issues 中 
https://github.com/gocrane/crane/issues/754
https://github.com/gocrane/crane/issues/757
bug 的修复 

详细描述:
 折扣这里 会存在负数的问题


pkg/web/src/modules/editClusterSlice.ts
```diff
@@ -1,3 +1,9 @@
+/*
+ * @Description: 
+ * @Author: 若城
+ * @Date: 2023-05-04 22:20:52
+ * @LastEditTime: 2023-05-04 22:23:49
+ */
```
qmhu, line 6: Please remove this

pkg/web/src/modules/editClusterSlice.ts
```diff
@@ -22,7 +28,7 @@ const defaultCluster = {
   id: v4(),
   clusterName: '',
   craneUrl: '',
-  discount: 100,
+  discount: 1,
```
qmhu, line 31: Why set the discount to 1? default value is 100.

pkg/web/src/modules/editClusterSlice.ts
RuoChengF: 好的

pkg/web/src/modules/editClusterSlice.ts
RuoChengF: 折扣的取值范围是 1-9.9

pkg/web/src/modules/editClusterSlice.ts
RuoChengF: 已经去掉了

pkg/web/src/modules/editClusterSlice.ts
zsnmwy: The range of 1-9.9 of the value is not correct.
The value will be passed to pql(Prometheus Query Language) string.

image

https://github.com/gocrane/crane/blob/087dc98230a5654abb556d5592f600983adc0f69/pkg/web/src/hooks/useCraneDiscount.ts#L4-L10

https://github.com/gocrane/crane/blob/087dc98230a5654abb556d5592f600983adc0f69/pkg/web/src/pages/Dashboard/Base/components/TopPanel.tsx#L10-L67

RuoChengF: 也就是取值范围 不能小于0 就可以了是吧 ,  还是说取值范围 在 0-100 之间

RuoChengF: 重新修复了一下  折扣不可为负数

mfanjie:
@RuoChengF can you make the PR description as English, especially make the commit message as English.

RuoChengF:
The Pr content submitted this time is to modify the discount input content shall not be lower than 0

pkg/web/src/pages/Settings/cluster/EditClusterModal.tsx
```diff
@@ -338,8 +338,12 @@ export const EditClusterModal = React.memo(() => {
                       value={cluster.discount}
                       onChange={(value: InputValue) => {
                         let discount = value;
+                        let reg = /^\d+(\.{0,1}\d+){0,1}$/
```
qmhu, line 341: Not a good way to implement it.

RuoChengF: Of course, this is definitely not a good way to implement it. There are two main reasons why I write regular expressions directly inside components.
1. The web side has a low utilization rate for its own encapsulation method, and it seems to have multiple files, which is rather messy, which is one of the reasons why I put forward the improvement of "READE.MD" in the issue
2. My implementation idea is based on the current code style. If I repackage this thing into a method for import verification, the amount of code generated is too much.

PR #785, opened 2023-05-10
perf: optimize recommender manager

lbbniu:
#### What type of PR is this?
optimize recommender manager

Fixes #

### DIFF
--- a/pkg/recommendation/recommender/hpa/registry.go
+++ b/pkg/recommendation/recommender/hpa/registry.go
@@ -21,12 +21,16 @@ type HPARecommender struct {
 	MaxReplicasFactor       float64
 }
 
+func init() {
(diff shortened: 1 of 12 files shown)


pkg/recommendation/recommender/hpa/registry.go
```diff
@@ -70,7 +74,7 @@ func NewHPARecommender(recommender apis.Recommender, recommendationRule analysis
 	}
 
 	return &HPARecommender{
-		*replicasRecommender,
+		*(replicasRecommender.(*replicas.ReplicasRecommender)),
```
qmhu, line 77: What's this line means?

lbbniu: NewReplicasRecommender return `recommender.Recommender` interface

Issue #845, opened 2023-08-08
upgrade cadvisor to reduce crane-agent cpu usage


xrmzju:
**Describe the bug**
The current version of cAdvisor (v0.39.2) reads extra cgroup files even though we have specified that the only metrics we need are [CpuUsageMetrics and ProcessSchedulerMetrics](https://github.com/gocrane/crane/blob/1da8f5e4900edb3e547dbaccf77771acd4cbcfcd/pkg/ensurance/collector/cadvisor/cadvisor_linux.go#L70) .This causes a lot of file reading, which consumes a lot of CPU. More details can be found in [this issue](https://github.com/google/cadvisor/issues/2932).

**Reproduce steps**

1. Run crane-agent on a node with more than 100 pods.
2. Observe the CPU usage, which can be quite high.
3. In our environment, running crane-agent on a node with 150 pods can consume up to 8 cores.

**Expected behavior**
We expect cAdvisor to only read the cgroup files that are necessary to collect the specified metrics, and not to read any extra files that are not needed.

**Screenshots**

**Environment (please complete the following information):**
 - K8S Version: [e.g. 1.22]

PR #846, opened 2023-08-08
upgrade cadvisor to reduce cpu usage

xrmzju:
#### What type of PR is this?
bugfix

upgrade cadvisor to v0.41.0 to reduce cpu usage

Fixes https://github.com/gocrane/crane/issues/845


xrmzju:
WeCom20230808-110755@2x

borgerli:
# My Test Results

## Scenario
on 16c/32G node, create 200 pods of nginx deployment and a svc for nginx pods. Use `ab -n 999999999 -c 500 http://<nginx-svc-ip>/` to do stress.
##Without this PR
Before stress, crane agent avg. cpu usage: 398m
After, 857m
## With this PR
Before stress, 237m
After, 380m

Issue #632, opened 2022-11-30
if workload utilization rate reaches the target，can stop recommending for these workload？


justzx2011:
## Describe the feature
如果负载cpu、内存使用率已达标，能加一个配置，停止这部分推荐吗？

qixiang-liu:
If cpu/mem meets the recommended value, is it not recommended? If this is the case, the recommended cpu/mem value should set a buffer.

qixiang-liu:
Enjoy it !!!!!

qixiang-liu:
Confirm the requirements here. If the recommended value is the same as the original value, is the target recommended value not written?

PR #854, opened 2023-08-23
fix: Add the configuration to meet the load cpu and memory usage. Sto…

qixiang-liu:
…p the recommendation

#### What type of PR is this?
[good first issue](https://github.com/gocrane/crane/labels/good%20first%20issue)

Fix issues

Fixes # (https://github.com/gocrane/crane/issues/632)


qmhu:
I think we need to define what is a stopped existing recommendation.

qixiang-liu:
I understand that the load cpu and memory usage are compliant, that the recommended value is equal to the source value in running, and not written in annotations.

If so, there should be something wrong with the one I fixed. Is that what it means?

qixiang-liu:
The operation of overwriting an existing recommendation is complete

qixiang-liu:
> I think we need to define what is a stopped existing recommendation.


The operation of overwriting an existing recommendation is complete

PR #859, opened 2023-08-30
add ContainerResource source type for ehpa prediction

whitebear009:

Fixes #

### DIFF
--- a/pkg/controller/ehpa/predict.go
+++ b/pkg/controller/ehpa/predict.go
@@ -133,25 +133,12 @@ func (c *EffectiveHPAController) NewPredictionObject(ehpa *autoscalingapi.Effect
 
 	var predictionMetrics []predictionapi.PredictionMetric
 	for _, metric := range ehpa.Spec.Metrics {
-		var metricName string
-		//get metricIdentifier by metric.Type and metricName
-		var metricIdentifier string
-		switch metric.Type {
-		case autoscalingv2.ResourceMetricSourceType:
-			metricName = metric.Resource.Name.String()
-			metricIdentifier = utils.GetMetricIdentifier(metric, metric.Resource.Name.String())
-		case autoscalingv2.ExternalMetricSourceType:
-			metricName = metric.External.Metric.Name
-			metricIdentifier = utils.GetMetricIdentifier(metric, metric.External.Metric.Name)
-		case autoscalingv2.PodsMetricSourceType:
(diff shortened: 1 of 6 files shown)


whitebear009:
@qmhu PTAL, thanks

pkg/controller/ehpa/predict.go
```diff
@@ -169,7 +156,23 @@ func (c *EffectiveHPAController) NewPredictionObject(ehpa *autoscalingapi.Effect
 						klog.Errorf("Got MetricRulesResource prometheus-adapter-resource Failed MetricName[%s]", metricName)
 					} else {
 						klog.V(4).Infof("Got MetricRulesResource prometheus-adapter-resource MetricMatches[%s] SeriesName[%s]", metricRule.MetricMatches, metricRule.SeriesName)
-						nameReg = utils.GetPodNameReg(ehpa.Spec.ScaleTargetRef.Name, ehpa.Spec.ScaleTargetRef.Kind)
+						podNameReg = utils.GetPodNameReg(ehpa.Spec.ScaleTargetRef.Name, ehpa.Spec.ScaleTargetRef.Kind)
+					}
+				}
+			case autoscalingv2.ContainerResourceMetricSourceType:
```
qmhu, line 162: Would you help to provide a sample and explain how to convert it? Also please show how to config it in prometheus-adapter mode. tks.

whitebear009: Here is a sample:
`  metrics:
  - type: ContainerResource
    containerResource:
      container: app
      name: cpu
      target:
        type: Utilization
        averageUtilization: 50
`
the default expression will convert it to promql like `irate(container_cpu_usage_seconds_total{container!="POD",namespace="%s",pod=~"%s",container="%s"}[%s])`  in file https://github.com/gocrane/crane/pull/859/files#diff-58fd591c9f408018596b9c5b19515f115c97a0243aeadcab4984c3d5bed6edc9R128

In prometheus-adapter mode, its config is basicly the same with pod resource metrics. The only difference is that compared to pod resource metrics, there is an additional field representing container name. This field name defaults to "container" and can be modified through the containerLabel of prometheus-adapter(https://github.com/kubernetes-sigs/prometheus-adapter/blob/master/pkg/config/config.go#L110) (this is also the original purpose of this field)

qmhu (approved):

Issue #876, opened 2023-11-27
Optimize pod qos initializer.


payall4u:
## Describe the feature

Offline pod with no PodQOS matched should be skip by qos initializer webhook.

payall4u:
Merged.

PR #878, opened 2023-11-29
Optimize qos initializer.

payall4u:
#### What type of PR is this?
Feature.

Optimize qos initializer webhook. Inject init-container and init-volume only when pod matched by offline pod qos.

From https://github.com/gocrane/crane/issues/876

### DIFF
--- a/pkg/ensurance/analyzer/analyzer.go
+++ b/pkg/ensurance/analyzer/analyzer.go
@@ -6,8 +6,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/gocrane/crane/pkg/ensurance/util"
-
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -29,6 +27,7 @@ import (
 	ecache "github.com/gocrane/crane/pkg/ensurance/cache"
 	"github.com/gocrane/crane/pkg/ensurance/executor"
 	"github.com/gocrane/crane/pkg/ensurance/executor/podinfo"
+	"github.com/gocrane/crane/pkg/ensurance/util"
 	"github.com/gocrane/crane/pkg/known"
 	"github.com/gocrane/crane/pkg/metrics"
 	"github.com/gocrane/crane/pkg/utils"
--- a/pkg/ensurance/analyzer/analyzer.go
+++ b/pkg/ensurance/analyzer/analyzer.go
@@ -6,6 +6,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/gocrane/crane/pkg/ensurance/util"
+
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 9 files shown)


tools/initializer/resource.yaml
```diff
@@ -100,28 +100,36 @@ data:
     kind: QOSConfig
     qosInitializer:
       enable: true
-      selector: 
+      selector:
         matchLabels:
           app: nginx
       initContainerTemplate:
-        name: crane-qos-initializer
+        name: qos-initializer-container
```
mfanjie, line 107: nit: name: qos-initializer is fine

tools/initializer/resource.yaml
```diff
@@ -100,28 +100,36 @@ data:
...
         image: docker.io/gocrane/qos-init:v0.1.6
         imagePullPolicy: IfNotPresent
+        args:
+          - "while ! grep -q gocrane.io/cpu-qos /etc/podinfo/annotations; do sleep 1; done; echo cpu qos setting competed;"
```
mfanjie, line 111: if it's just a shell call, do we still need qos-init image, can it be changed to pause?

pkg/webhooks/pod/mutating.go
```diff
@@ -28,7 +40,7 @@ func (m *MutatingAdmission) Default(ctx context.Context, obj runtime.Object) err
 		return fmt.Errorf("expected a Pod but got a %T", obj)
 	}
 
-	klog.Infof("Into Pod injection %s/%s", pod.Namespace, pod.Name)
+	klog.Infof("mutating started for pod %s/%s", pod.Namespace, pod.Name)
```
mfanjie, line 43: A code standard, all logs start with Capital letters, and please check if a log level should be added and disabled the msg by default. same suggestion to all newly added logs

pkg/webhooks/pod/mutating.go
```diff
@@ -47,17 +59,54 @@ func (m *MutatingAdmission) Default(ctx context.Context, obj runtime.Object) err
 		return err
 	}
 
-	if ls.Matches(labels.Set(pod.Labels)) {
-		if m.Config.QOSInitializer.InitContainerTemplate != nil {
-			pod.Spec.InitContainers = append(pod.Spec.InitContainers, *m.Config.QOSInitializer.InitContainerTemplate)
+	if !ls.Matches(labels.Set(pod.Labels)) {
+		klog.Infof("injection skipped: webhook is not interested in the pod")
```
mfanjie, line 63: injection->Injection, and this is a debug log, I suggest to add a log level and do not enable by default, and if we have label config for the webhook, this line of code should not be called?

payall4u: In fact, I don't like this logic. But in order to be compatible with the previous design, I still leave the matching here.

pkg/ensurance/analyzer/analyzer.go
```diff
@@ -6,6 +6,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/gocrane/crane/pkg/ensurance/util"
```
mfanjie, line 9: move this to the crane import section

pkg/webhooks/pod/mutating.go
```diff
@@ -4,12 +4,16 @@ import (
 	"context"
 	"fmt"
 
-	corev1 "k8s.io/api/core/v1"
+	"github.com/gocrane/crane/pkg/ensurance/util"
```
mfanjie, line 7: org import

mfanjie (approved):
LGTM

PR #881, opened 2023-12-28
tsp indicator isolated from ehpa

whitebear009:
#### What type of PR is this?
optimize

Currently the tsp indicator is calculated in ehpaList. If the tsp is created separately (not through ehpa), then the indicators of this tsp will be gone.

Fixes #


whitebear009:
@qmhu  PTAL, thx

PR #900, opened 2024-04-16
 fix issues #898 

### LINKED ISSUE #898: craned cpu utilization continues at 100% and resource and replicas recommended values are older
**Describe the bug**
When using craned for Resource and Replicas recommendation in a larger cluster, when the number of workloads reaches 2w+, the CPU utilization continues to be 100% (limit 8c), resource and replicas recommended values are older( runInterval: 24h0m0s)
**Reproduce steps**
1. Create more than 20,000 workload objects;
2. Configure recommended rules for all workloads;

**Expected behavior**
The Resource and Replicas recommendation work properly

**Screenshots**
image
image

 - K8S Version: [e.g. v1.14.3]
 - Crane Version: [e.g. pr-872-bb5c1e4]

Cloudzp:

Fixes #898


pkg/controller/recommendation/recommendation_rule_controller.go
```diff
@@ -528,3 +536,106 @@ func IsConvertFromAnalytics(recommendationRule *analysisv1alph1.RecommendationRu
...
+	return &RecommendationIndex{
+		idx: idx,
+	}
+}
+
+func createIndexKey(r *analysisv1alph1.Recommendation) IndexKey {
+	return IndexKey{
+		Kind:        r.Spec.TargetRef.Kind,
+		APIVersion:  r.Spec.TargetRef.APIVersion,
+		Namespace:   r.Spec.TargetRef.Namespace,
```
mfanjie, line 624: confirmed with @Cloudzp, no matter what mode craned is running, namespace is always being set on recommendation obj. so this safe for now, we need to make sure namespace is always being set even recommendation and target is in same namespace, otherwise this would cause 
messes

mfanjie (approved):
LGTM

PR #910, opened 2024-08-22
Cherrypick to support v1.26

Cloudzp:
Cherry-pick some craned-related commits to support the v1.26 branch

Fixes #


abrahamhwj:
metricAdapter startup error, indicating “unknown flag: --alsologtostderr", and also "--v"
Who can help answer it, thx very much：
1. What is the reason for this problem？
2. What do these two flags do?
3. What's the problem if cancel them?

yangyegang:
metricAdapter 和 craned  startup error：“unknown flag: --alsologtostderr", and also "--v"，这个有解决吗？

PR #915, opened 2024-12-04
fix the issue that trigger recommendation by run number abormal

Cloudzp:
#### What type of PR is this?
fix the issue that trigger recommendation by run number abormal

Fixes #


mfanjie (approved):
LGTM

PR #923, opened 2025-05-14
support recommend with pod metric

github-wyy:
#### What type of PR is this?
support recommend with pod metric

Fixes #


XuShaohui (approved):
lgtm

